PR #36, opened 2023-11-14
Fix config.json issues

### LINKED ISSUE #33: Various config.json issues
### Unlisted dependency on `jq`

Hi, the primary reason I filed this under "bug" is that I noticed to use the `config.json` file in the first place requires package `jq` to be installed on the user's system but this is not listed as a dependency for the project anywhere. I didn't have this installed and that was hard to pinpoint :smile: 

So for example, this package is a Vivify dependency for me: https://archlinux.org/packages/extra/x86_64/jq/

Maybe the readme should have a Dependencies -section, and do you know if there could be more?

### `"pageTitle"` doesn't seem to work with `basename` or `join`

If my goal is to set the pagetitle as `basename(path)`, what should I set on the key? I figured it should be this:

```
{
    "pageTitle": "basename(path)"
}
```

However each time I try this, using `viv` gives an "Unable to connect" firefox page.

What has worked as a `"pageTitle"` key:

* `"'foo'"` will show <title> as `foo`
* `"console.log('foo')"` will show <title> as `undefined` (correct, I think :smile: )
* `"path"` will show the full path, which is a good sign
  * Also: `"console.log(path)"` correctly shows `undefined`

Same issue (Unable to connect) applies for trying to use the `join` func, as an example:

* `"join('foo', path)"`

This is what the Firefox page looks like:

image

jannis-baum:
Close #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ issue](https://github.com/jannis-baum/vivify/issues/new/choose) or
 
 ### Config
 
-Vivify will look for a config file at `~/.vivify/config.json` and
+Vivify will look for an optional config file at `~/.vivify/config.json` and
 `~/.vivify.json`. This file should contain a JSON object that can have the
 following optional keys:
 
@@ -64,6 +64,9 @@ following optional keys:
   JavaScript code that will be evaluated to determine the
   viewer's page title based on the variable `path` for the given file (default
   is `join(basename(dirname(path)), basename(path))`, e.g. `my_dir/my_file`)
+
+Note that you need to have [`jq`](https://github.com/jqlang/jq) installed if you
+want to use a custom config file.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -61,9 +61,16 @@ following optional keys:
   }
   ```
 - **`"pageTitle"`**\
-  JavaScript code that will be evaluated to determine the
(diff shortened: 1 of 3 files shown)


tuurep:
Great, thank you! It's a fix, got my preferred title with either:

* "\`${basename}\`"
* "basename"

Tried something like:

* "\`${basename} in ${dirbasename}\`"

and works as expected. JSON's mandatory doublequotes can be a bit awkward but that's just JSON.

I couldn't however find where the logging is output on a title error, nothing in either firefox dev console or terminal. But it's a big improvement, still rendering the page and knowing there was a title error.

tuurep (approved):
I think this is fine, but if you're concerned about me not finding the logs, let me know.

jannis-baum:
> I think this is fine, but if you're concerned about me not finding the logs, let me know.

Okay, just confirmed the problem. The way I'm doing it now works with Safari but not Firefox. Will adjust later & then let you know :)

jannis-baum:
@tuurep I switched to writing the error message straight onto the page, what do you think?

tuurep:
Yes, this works great. Thanks & ready to merge

image

^ sorry I don't know what this means, I figured it's probably nothing but if it requires something from me let me know :smile:

PR #39, opened 2023-11-19
Filename can't contain non-ascii character (such as `ä`, `é`)

### LINKED ISSUE #26: Filename can't contain non-ascii character (such as `ä`, `é`)
### Describe the bug

I had a file with letter `ä` in the name. Trying to open this with `viv` opens a `File not found` page.

The address field shows the correct name for the file.

The tab title shows `ä` as `%C3%A4` initially, but if I move the tab around, it then shows the correct name.

### To Reproduce

Vivify version `0.0.3`, browser Firefox

1. Create a file with the name `café.md`
2. Write contents `Foo Bar QuuuX`
3. Open file with `viv café.md`

### Expected behavior

File should be found

### Additional context

I guess it's debatable whether filenames are allowed to contain non-ascii characters.

Also can't handle spaces in filenames (while I don't name files this way myself, I often see it in the wild for whatever reason). Example `Foo Bar.md`

jannis-baum:
Close #26

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,3 +8,25 @@ sure you follow the steps below:
   with `#<your issue number>` as the context for all commits
 - Make sure you are not adding any merge commits to your branch
 - Open a pull request & wait for a review!🩵
+
+## Developing environment
+
+After installing all dependencies with `yarn`, you can run
+
+```sh
+yarn dev
+```
+
+to run the development server for Vivify on port 3000 that will (1)
+automatically reload when you make changes to the code, and (2), unlike the
+installed version, not shut down when there are no connected clients.
+
+With the development server running, use
+
+```sh
+yarn viv
+```
+
+instead of your installed Vivify executable. This will (1) connect to the
+development server on port 3000 instead of running the installed server, and (2)
+use the `viv` executable in the repository.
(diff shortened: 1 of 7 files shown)


jannis-baum:
Hey @tuurep I did my best to test lots of scenarios and check that everything works but it's very possible that I missed something ^^ I would appreciate if you could test this as well & leave a review! Thanks!

---

As a heads up, there is one caveat that I have found: image paths in markdown files can't contain spaces because our markdown parser doesn't recognize it as an image anymore if the path has a space. But IMO this is a pretty far out edge case that I feel comfortable blaming on the user and it would probably be tedious to tend to because we would have to look into the parser. So I'll ignore it ^^

jannis-baum:
Oh, I also added `yarn viv` to make the dev setup a bit easier (see `CONTRIBUTING.md`). So now I just do `yarn dev` to start the server & then `yarn viv` to open stuff through the dev server

tuurep:
Thanks, I'll be testing this within like 12 hours :)

tuurep:
Yeah this is working great! Thanks a lot, I can't find any issues with it for now.

Also finding the `yarn dev` & `yarn viv` useful :D sometimes was getting tiresome especially with the `killall vivify-server` I was doing after every code change. Made sure the `yarn dev` didn't need to be rebooted after change and yep, works!

Yeah the imgpath sounds good, sounds like it's out of our hands (or should be) :smile:

tuurep (approved):

jannis-baum:
Nice thank you!☺️

Issue #38, opened 2023-11-18
Slightly indented first line of a `pre` code block


tuurep:
![image](https://github.com/jannis-baum/vivify/assets/15036238/789d979d-48b1-43fb-8963-48e4bf73f0e9)

Looking into this now!

Facts:
* Happens in every single `pre` block
* Not a space character `" "` but an alignment issue

jannis-baum:
Nice! After this and the light mode are done we can [release `0.1.0`](https://github.com/jannis-baum/vivify/milestone/1). Or was there something else you wanted to include before?

tuurep:
Yeah that's good!

I'm still interested in the `ä` in filenames issue, but that can be later (also I can't tell now if that's complicated or not),

and making the `AUR` package, but that can be a separate thing.

jannis-baum:
> I'm still interested in the ä in filenames issue, but that can be later (also I can't tell now if that's complicated or not)

I'll say this really carefully and with a huge grain of salt but I think that should be pretty straight forward🙈 Maybe I'll find time tomorrow or early next week to try fixing it. If I don't make it before you finish the other two open issues then that can be 0.1.1 ^^

> and making the AUR package, but that can be a separate thing.

Oki once you want to start working on it, open an issue & then we can talk there. I have no idea how that works but in case we need any additional CI/CD related stuff for it I can help!

tuurep:
> I'll say this really carefully and with a huge grain of salt but I think that should be pretty straight forward🙈 Maybe I'll find time tomorrow or early next week to try fixing it. If I don't make it before you finish the other two open issues then that can be 0.1.1 ^^

Sounds great!

> Oki once you want to start working on it, open an issue & then we can talk there. I have no idea how that works but in case we need any additional CI/CD related stuff for it I can help!

Yeah, I can totally take responsibility of the whole thing since to make sure the package works you'd have to run Arch Linux :D but 2 things: 

1. If you want to be co-maintainer of the package just ask
2. I can let you review the `PKGBUILD`s and there might be like a `yarn` / `tsc` build step or something like that which I'm a little unfamiliar with, and you can check if it looks correct. I can open an issue here for discussion later

### On the indent issue

Think I'm onto something!

https://github.com/jannis-baum/vivify/blob/68afc9b45880ba43d8f739e875ce13fa7c3bc7f9/src/parser/highlight.ts#L15

There if I remove the `<code>` tags, there's no indent issue although now the block has a bit of a left side padding like this:

[...]

PR #40, opened 2023-11-19
fix: slightly indented first line of pre block

tuurep:
Close: https://github.com/jannis-baum/vivify/issues/38

@jannis-baum what you said here looks like it worked perfect:
> let's fix it in the CSS by `unset`ting the padding of `code`s in `pre`s like we do with the background color.


jannis-baum (approved):
nice hehe

Issue #17, opened 2023-08-06
Light mode


jannis-baum:
auto-sync with system

tuurep:
Have you thought about a default light theme?

I tried to make one that looks as "normal" as possible, and ended up mostly copying Github's markdown colors:

https://github.com/tuurep/dotfiles/blob/master/.vivify/styles.css

I'm very happy with this, here's how Vivify's readme looks for example:

![image](https://github.com/jannis-baum/vivify/assets/15036238/73611c84-522f-432a-9faf-2ab3c62f1945)

My `code` font looked a little faint compared to GH so darkened the color of that

Blockquotes are also mimicking Github's style, which required me to set `margin-left` as 0

jannis-baum:
Looks great! I'd be happy about a PR that adds this to the existing style sheet using [CSS' media query `prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) so that Vivify's theme syncs with the system.

One thing you might need to check still is syntax highlighting; if I saw the screenshot correctly then that's not yet specified in your light theme and the fallback dark theme colors might not look good on bright background.

tuurep:
Thank you, I can send it if I can manage to do some tweaks.

Yeah in fact I explicitly set the syntax off in this way:

```
span[class^="hljs"] {
    color: #16191d;
    font-weight: normal;
}
```

... to avoid obsessing over syntax highlighting for the time being :smile: but for the PR I'll do it - the `hljs-` groups aren't too many and probably I can mimic Github's colors on those too.

I have a question though, about the blockquotes: should I leave that out (only set it in my personal configs), or do you think the Github style should be default?

For comparison, this is how it currently is:

![image](https://github.com/jannis-baum/vivify/assets/15036238/501897b0-9b71-4bcd-9604-20dbb8b957e1)

This is how I configured it:

![image](https://github.com/jannis-baum/vivify/assets/15036238/72bdb3e0-2ac7-4f99-a48c-86515cc94203)

CSS for the above:

``` css
blockquote {
    padding: 0 1em;
    color: #656d76;
    border-left: .25em solid #d0d7de;
    margin-left: 0
}
```

Dark theme would need to set a color for quote text and the left side border.

jannis-baum:
Sounds great! Thank you!!

[...]

PR #43, opened 2023-11-23
feat(#17): implement light mode

tuurep:
Issue: https://github.com/jannis-baum/vivify/issues/17

Okay got this to a great point to start! Raises a few questions:

### No configuration defaults

What should be the default lets say in a case where someone doesn't know what their `prefers-colorscheme` is?

With the changes in this PR's initial commit, I go into the `prefers-colorscheme light` branch, which is coincidentally exactly what I want :smile: However, I don't know what determines this, is it simply the case that if you don't explicitly set `dark`, the default is `light`?

Thinking if there's need for a config option something like `theme`: `[light|dark|system]`, but it's possible that it's not even necessary. What do you think?

### Maintenance of the github syntax theme

Got the syntax theme from here:

https://github.com/highlightjs/highlight.js/blob/main/src/styles/github.css

And as far as I saw, looks great! Looks like someone is "maintainer" of that in the highlight.js repo. Was thinking if since it's like a builtin of highlight.js, is there another way to get it from the lib/module than copypasting the CSS?

Anyways, it has not been updated for 2 years so it doesn't need much attention I think.

### Overflow: auto

I've changed the default of `pre` `overflow: scroll` to `overflow: auto`, because `overflow: scroll` looks pretty harsh on my browser in the case there's nothing to scroll - here's example screenshot:

image

with `auto`, in the above case there's no such thick "borders", but still get the benefit of being able to scroll only in the case it's needed. If there's a reason `overflow: scroll` is needed, let me know :)

### Other

See if the new dark theme blockquote colors work for you, and adjust as necessary.

For the light theme I've stripped out some highly personal choices, mainly with different link-type foreground colors (anchor link and the parent-dir link), so all links are just `a` foreground. Now I'm pretty happy with how this theme would work as a sensible default. Thoughts / criticisms?

### DIFF
--- a/static/highlight.css
+++ b/static/highlight.css
@@ -42,3 +42,121 @@
     color: #ffdcd7;
     background-color: #67060c;
 }
+
+/* light mode */
+@media (prefers-color-scheme: light) {
+    .hljs {
+        color: #24292e;
+        background: #ffffff;
+    }
+
+    .hljs-doctag,
+    .hljs-keyword,
+    .hljs-meta .hljs-keyword,
+    .hljs-template-tag,
+    .hljs-template-variable,
+    .hljs-type,
+    .hljs-variable.language_ {
+        /* prettylights-syntax-keyword */
+        color: #d73a49;
+    }
+
+    .hljs-title,
+    .hljs-title.class_,
+    .hljs-title.class_.inherited__,
+    .hljs-title.function_ {
+        /* prettylights-syntax-entity */
+        color: #6f42c1;
+    }
+
+    .hljs-attr,
+    .hljs-attribute,
+    .hljs-literal,
+    .hljs-meta,
+    .hljs-number,
+    .hljs-operator,
+    .hljs-variable,
+    .hljs-selector-attr,
+    .hljs-selector-class,
+    .hljs-selector-id {
+        /* prettylights-syntax-constant */
+        color: #005cc5;
+    }
+
+    .hljs-regexp,
+    .hljs-string,
+    .hljs-meta .hljs-string {
+        /* prettylights-syntax-string */
+        color: #032f62;
+    }
+
+    .hljs-built_in,
+    .hljs-symbol {
+        /* prettylights-syntax-variable */
+        color: #e36209;
+    }
+
+    .hljs-comment,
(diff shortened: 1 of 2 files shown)


jannis-baum:
Thank you, everything looks great! Also like the block quote in both modes & don't mind about the overflow thing. Sorry for taking a bit longer to review this time.

**tldr;** IMO this is good to merge, I'd just like to include the URL to the GitHub theme in the CSS :)

> Maintenance of the github syntax theme

Cool! I think this is fair, no need for any complicated setup to pull something that's not frequently updated. The katex CSS and LaTeX fonts are also copied straight into this repo. Would you mind adding the source URL into the code as a comment though?

> No configuration defaults

**Disclaimer** - the rest of this is a bit of a rant haha

IMO this is not Vivify's job. The way the CSS is set up now, the dark theme is default and it will change to light iff the user's browser reports `prefers-color-scheme: light`. The way I see it, this is the way it should be [^1]. If every program/app creates their own dark/light config we end up with a huge mess and a pain to switch themes. I think someone who isn't happy with the theme they get on Vivify should take care of their setup themselves, because in this case their issue is not only with Vivify but also with just about every other website that doesn't put (IMO) inappropriate switches into their own config options ^^

The only argument for having a config option that I see is in case someone for whatever reason wants Vivify to have a different theme than everything else, but if this doesn't affect you then I wouldn't add it until someone complains :) Especially because there is also the option to supply your own CSS

[^1]: aside from which theme comes up for (the probably very rare) browsers that don't report anything for `prefers-color-scheme`, but I don't really care about this to be honest🙈

jannis-baum (approved):

tuurep:
Nice! Fairly surprised I got everything good on the first go :) Added the URL comment.

Thanks for all the work, after this one we've pretty much fixed all of my personal gripes so I can just start using this for writing notes.

I'll probably look into making the AUR package some time this week.

jannis-baum:
> Thanks for all the work, after this one we've pretty much fixed all of my personal gripes so I can just start using this for writing notes.

Awesome! Thank you for all your work as well! Very valuable contributions!

> I'll probably look into making the AUR package some time this week.

Great!

PR #54, opened 2023-12-04
AUR package

### LINKED ISSUE #52: AUR package
Hi, I've managed to create the AUR package!

https://aur.archlinux.org/packages/vivify

Just now managed to install it using AUR helper [`yay`](https://github.com/Jguer/yay), here's what it looks like when querying with yay:

image

Very cool, works perfectly fine :)

You can take a look at the PKGBUILD [here](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=vivify) and if something looks wrong, let me know.

What I can think of is:
1. Project has no license, if you have one in mind it might be wise to add
2. The reason `gcc-libs` is in `depends` is just because it was raised in an error using [`namcap`](https://wiki.archlinux.org/title/namcap) to check the package, so I guess it should be in there and it's good to be safe with the dependencies

Other than that, it's dead simple when I just run your build script :smile:

Maybe we can add link on the README?

jannis-baum:
Close #52

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,20 @@ issue](https://github.com/jannis-baum/vivify/issues/new/choose) or
 - syntax highlighting for code
 - graphviz/dot graphs
 
+### Editor Support
+
+Vivify has a simple API to integrate your favorite editor so the viewer live
+updates to any changes as you are typing and the scrolling is smoothly
+synchronized!
+
+See below for a list of existing editor plugins. In case your favorite editor is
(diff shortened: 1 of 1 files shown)


jannis-baum:
@tuurep Let me know if you're happy and then I'll merge & release☺️

tuurep:
Yeah! That's a good wording, thank you :)

Ready for merges

jannis-baum:
Release is in progress!

Issue #57, opened 2024-01-02
Github Flavored Markdown extended syntax support


tuurep:
Github Flavored Markdown syntax extensions:

- [x] Tables (style added in #58)
- [x] Task list
- [x] Strikethrough
- [x] Autolinks (added in #58)
- [ ] Footnotes (#59)

jannis-baum:
Nice, thank you so much, this is great!

> Is it true that fixing the Extended Syntax that doesn't work depends mostly on the library markdown-it and what it supports?

Yes! Supporting any part of the extended syntax will come down to finding a good and well-maintained plugin for `markdown-it` and integrating it into Vivify.

> I'll definitely fix the style issues as soon as possible (tables, horizontal rule)

Thank you!!

> I would at least like Footnotes to be supported in Vivify.

I agree.

All the other things you mentioned that GitHub doesn't support I would argue we don't need either? Or at least I don't haha[^1]. But if your goal is for Vivify to support all extended syntax then I'll definitely support that :)

[^1]: Aside from footnotes the only thing here that I tend to use sometimes but very rarely are definition lists. When I use those they also need to be rendered on GitHub though so I usually opt for HTML.

tuurep:
Hmmm... let's see, let's at least do these:

1. Tables style
2. Horizontal rule style
3. Footnotes

In case it would be simple and easy, of course would be nice to support all, but basically I agree. Interesting that you've used 'definition lists', I've never known about it and the markdown syntax for it would be really comfortable, but yeah since it doesn't render in github a lot of the usefulness is gone there.

What do you think about Automatic URL Linking though? That's something I'd likely use myself, but it's not a big deal to wrap links in `[]()` instead

jannis-baum:
> What do you think about Automatic URL Linking though? That's something I'd likely use myself, but it's not a big deal to wrap urls in []() instead.

Yes this sounds good! Especially since GitHub does it as well. It's very easy to do natively with `markdown-it`, I'll add it to your PR :)

jannis-baum:
Oh by the way, take a look at [this](https://www.npmjs.com/package/markdown-it#syntax-extensions); for lots of the main extended features like footnotes they recommend a plugin at `markdown-it` already. So let's use the recommendation where it exists

tuurep:
> > What do you think about Automatic URL Linking though? That's something I'd likely use myself, but it's not a big deal to wrap urls in  instead.
> 
[...]

PR #59, opened 2024-01-10
feat(#57): footnotes

tuurep:
Addresses the last task in #57 

Can you make sure I added the yarn module in the right way? I ran `yarn add markdown-it-footnote`

Are these styles good? I straight up added `20px` margin-top for the footnotes section. Dark mode needed a new color which is called something like `fg-dimmed` in a variable in Github's styling.

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,9 +26,10 @@
     "markdown-it": "^13.0.1",
     "markdown-it-anchor": "^8.6.7",
     "markdown-it-emoji": "^2.0.2",
+    "markdown-it-footnote": "^4.0.0",
     "markdown-it-inject-linenumbers": "^0.3.0",
-    "markdown-it-texmath": "^1.0.0",
     "markdown-it-task-lists": "^2.1.1",
+    "markdown-it-texmath": "^1.0.0",
     "uuid": "^9.0.0",
     "ws": "^8.13.0"
   },
--- a/src/parser/parser.ts
+++ b/src/parser/parser.ts
@@ -20,6 +20,7 @@ mdit.use(anchor, {
 /* eslint-disable @typescript-eslint/no-var-requires */
 mdit.use(require('markdown-it-emoji'));
 mdit.use(require('markdown-it-task-lists'));
+mdit.use(require('markdown-it-footnote'));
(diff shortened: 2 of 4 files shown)


jannis-baum (commented):
Nice, looks great! I'll have a proper look tomorrow :)

For now, [I replied to your question in the issue](https://github.com/jannis-baum/vivify/issues/57#issuecomment-1885442451) so maybe let's use a margin and font size in `em`/`rem` units? Makes it more accessible for people who need bigger text, for example. There are probably more spots where we should switch to relative units but I guess that's a problem for the future🙈

tuurep:
Ok! I'll commit once I can figure out good values for font and margin

tuurep:
Ah I figured it out: `0.75rem` and `1.25rem`

Yeah the idea to convert raw `px` values to `rem` seems good. Though I was wondering about something like: why are some of the border elements exactly `0.2px` tall and is there even a difference between `0.2px` and `1px`

But we can merge this and think more broadly, if you want

jannis-baum:
> Yeah the idea to convert raw `px` values to `rem` seems good. Though I was wondering about something like: why are some of the border elements exactly `0.2px` tall and is there even a difference between `0.2px` and `1px`

Good question. I just learned about it from [w3.org](https://www.w3.org/Style/Examples/007/units.en.html):

> The px unit is the magic unit of CSS. It is not related to the current font and usually not related to physical centimeters or inches either. The px unit is defined to be small but visible, and such that a horizontal 1px wide line can be displayed with sharp edges (no anti-aliasing). What is sharp, small and visible depends on the device and the way it is used: do you hold it close to your eyes, like a mobile phone, at arms length, like a computer monitor, or somewhere in between, like an e-book reader? The px is thus not defined as a constant length, but as something that depends on the type of device and its typical use.

jannis-baum (approved):
Thank you!!

jannis-baum:
Hey @tuurep just a quick out-of-place question🙈 Is there anything else related to styles or GitHub extended markdown features you want to add/change? If not I would make a new minor release now that this is merged

tuurep:
Yeah I have nothing to add for now, and great, I could benefit from a release right now :) Thanks

PR #90, opened 2024-07-14
ESM and simplified config

jannis-baum:
This PR makes two bigger changes:

1. It changes Vivify to be an [ES Module](https://nodejs.org/api/esm.html) and no longer CommonJS: Seems to be a trend that people are moving to ESM. Using ESM-only dependencies from CommonJS is a pain. We had a whole very ugly file just to use the ANSI-parser before because it was ESM-only and for this PR I had to adopt a second ESM-only ([open](https://www.npmjs.com/package/open)) dependency so I decided to make the switch.
   - the build setup now uses `tsc` to compile to JavaScript before bundling with `webpack` because I couldn't get it to work in one go anymore (though it also seems more understandable now)
   - imports of local files now have a `.js` suffix
   - the importing of non-typed markdown-it modules got a bit more annoying
   - some other things got simplified quite a bit, most notably ESM dependencies
3. The config is now handled entirely by `vivify-server`, making it a lot more clear and manageable, and `viv` gets simplified a lot in that we no longer need `jq` or an `open`/`xdg-open`-command.


jannis-baum:
@tuurep I would appreciate if you could review this as well after #72 is done! But no worries if it's too much work. Just let me know :)

tuurep:
I don't know much about CommonJS and/or ESM but I made sure stuff builds - :+1: 

Then tested my configs still apply - :+1:

But when running I notice one big problem:

I can no longer open more than one viewer at a time, example:

```
$ viv README.md            # Opens viewer on the browser
$ viv CONTRIBUTING.md # Does nothing!
$ killall vivify-server
$ viv CONTRIBUTING.md # Now I can open this, but nothing else at the same time
```

jannis-baum:
Awesome, thanks!

> But when running I notice one big problem:
>
> I can no longer open more than one viewer at a time, example:

This is strange👀 I just tried it again on mine and can't reproduce it unfortunately, for me it works. Could you try debugging it? Suggestions:

1. Ensure you have the current `viv` and not the old one
2. Try calling `vivify-server` directly. `viv` is just a wrapper now to start and `nohup` the `vivify-server`, so you could do `vivify-server README.md` (this should start the server and open the README, keeping the terminal blocked because the server is running there), and then `vivify-server CONTRIBUTING.md` (this should just open the CONTRIBUTING.md and then exit)
3. If 2. has the same behavior, check/print-debug what happens in the last 24 lines of `src/app.ts`; these are responsible for the opening and should be at fault then.\
   If 2. works, then `viv` is at fault and we have to look into it there.

tuurep:
I'll suggest this on a hunch before starting to debug properly:

Could it be an asynchronity thing?

```
get(`${address}/health`, () => {
    // server is already running
    openArgs();
    process.exit(0);
}).on('error', () => {
    // server is not running so we start it
    server.listen(config.port, async () => {
        console.log(`App is listening on port ${config.port}!`);
        openArgs();
    });
});

```

Something executing out of order and stuff like that? I always make these mistakes when I deal with these.

tuurep:
Ah, my fork doesn't have the Actions the same way this repo does. Lets see how I'd get the build artifact (just trying to add some debug console logs).

tuurep:
Although hold on. Now is an appropriate time to check the `nvm` thing [here](https://github.com/jannis-baum/vivify/discussions/80)

tuurep:
So I added a test console log in the `// server is already running` block and it always logs, but appears as if `process.exit(0)` is done before `openArgs()`

tuurep:
More logging reveals it reaches

```
open(url);
```

in `openArgs()`

but executes `process.exit(0);` before `open(url);` which looks like it's async: https://www.npmjs.com/package/open

src/app.ts
```diff
@@ -30,23 +24,43 @@ app.use('/viewer', viewerRouter);
...
+        const absolute = presolve(path);
+        const url = `${address}${pathToURL(absolute)}`;
+        open(url);
+    }
+};
+
+get(`${address}/health`, () => {
+    // server is already running
+    openArgs();
+    process.exit(0);
```
tuurep, line 59: `process.exit(0)` needs to wait that `openArgs()` is finished (specifically `open(url)` if I've understood correctly)

tuurep (changes requested):

src/app.ts
jannis-baum: Oh, I didn't realize this was async. My bad. Will fix it later.

jannis-baum: Can you check if it works now? I hopefully fixed it in https://github.com/jannis-baum/vivify/pull/90/commits/b72355a7ca374e7fd362d67953b5d68dc6eb1e72 :)

tuurep: Yes, fixed! Thank you

tuurep (approved):

tuurep:
Ok! Are we ready to merge?

PR #78, opened 2024-07-12
New config options

jannis-baum:
Close #35

PR #98, opened 2024-07-15
Fix running CI on PRs

### LINKED ISSUE #97: Fix running CI on PRs

jannis-baum:
Close #97

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,6 +6,9 @@ on:
       - '**'
     tags:
       - 'v*'
+  pull_request:
+    branches:
+      - main
 
 permissions:
   contents: write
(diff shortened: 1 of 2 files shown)


jannis-baum:
Hey @tuurep I added this yesterday so that the CI also runs on PRs coming from forks. We had this before but then in the context of switching to SEA we changed it to run on "all branches" so that there would always be access to the CI build. "All branches" only includes branches on the main repo, not on forks, so having only that isn't enough.

Having both however now seems to make the CI run twice on open PRs from branches on the main repo (e.g. for https://github.com/jannis-baum/vivify/pull/78 every action ran twice) which is quite wasteful.

Long story short, if I read along correctly you can now build yourself again, right? I.e. can we remove the `on: branch` target again so that the CI only runs on PRs to main (and main itself)?

tuurep:
> I.e. can we remove the `on: branch` target again so that the CI only runs on PRs to main

Yeah I can build now using `nvm` which works consistently so this is fine by me

PR #104, opened 2024-07-17
Support full extended syntax

### LINKED ISSUE #87: Support full extended syntax
The only things missing are super/subscripts, "highlight" and custom heading ids: https://www.markdownguide.org/extended-syntax/

- [x] add markdown-it extensions for these
- [x] simplify the README and just say we support full extended syntax
- [x] add features to `markdown-extended` testing file

jannis-baum:
Close #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,13 @@ issue](https://github.com/jannis-baum/vivify/issues/new/choose) or
 - links to other files: [relative links like in
   GitHub](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#relative-links)
   as well as absolute file links
+- [add styles, classes, ids or other attributes directly from
+  Markdown](https://github.com/arve0/markdown-it-attrs?tab=readme-ov-file#examples)
+
+You can find examples for all supported features in the files in the
+[`tests/`](tests) directory. In case you are looking at these on GitHub, keep in
+mind that GitHub doesn't support some of the features that Vivify supports so
+some things may look off.
 
 ### Editor Support
 
--- a/tests/markdown-additional.md
+++ b/tests/markdown-additional.md
@@ -28,3 +28,7 @@ digraph {
 While not a markdown syntax, this has a default style:
 
 Press <kbd>Ctrl</kbd> + <kbd>C</kbd> to copy, and <kbd>Ctrl</kbd> + <kbd>V</kbd> to paste!
+
+## Custom attributes
+
+This paragraph has a red background color.{style=background-color:red}
--- a/tests/markdown-extended.md
+++ b/tests/markdown-extended.md
@@ -89,9 +89,9 @@ That is so funny! :joy:
 
 :cowboy_hat_face::nerd_face:
 
-## Highlight
+## Highlight/Mark
 
-I need to highlight these ==very important words==.
+I need to highlight/mark these ==very important words==.
 
 ## Subscript
 
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "markdown-it-emoji": "^3.0.0",
     "markdown-it-footnote": "^4.0.0",
     "markdown-it-inject-linenumbers": "^0.3.0",
+    "markdown-it-mark": "^4.0.0",
     "markdown-it-sub": "^2.0.0",
     "markdown-it-sup": "^2.0.0",
     "markdown-it-task-lists": "^2.1.1",
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "katex": "^0.16.11",
     "markdown-it": "^14.1.0",
     "markdown-it-anchor": "^9.0.1",
+    "markdown-it-attrs": "^4.1.6",
     "markdown-it-deflist": "^3.0.0",
     "markdown-it-emoji": "^3.0.0",
     "markdown-it-footnote": "^4.0.0",
--- a/package.json
+++ b/package.json
@@ -21,6 +21,8 @@
     "markdown-it-emoji": "^3.0.0",
     "markdown-it-footnote": "^4.0.0",
     "markdown-it-inject-linenumbers": "^0.3.0",
+    "markdown-it-sub": "^2.0.0",
+    "markdown-it-sup": "^2.0.0",
     "markdown-it-task-lists": "^2.1.1",
     "markdown-it-texmath": "^1.0.0",
     "node-stream-zip": "^1.15.0",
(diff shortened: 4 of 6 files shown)


jannis-baum:
Hey @tuurep I added the features that were still missing for full extended syntax support. Was super fast, the features were all added by just adding the respective plugins. I think the crux in reviewing is to check if you are okay with the stuff I wrote on the README🙈

README.md
```diff
@@ -32,6 +32,13 @@ issue](https://github.com/jannis-baum/vivify/issues/new/choose) or
 - links to other files: [relative links like in
   GitHub](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#relative-links)
   as well as absolute file links
+- [add styles, classes, ids or other attributes directly from
+  Markdown](https://github.com/arve0/markdown-it-attrs?tab=readme-ov-file#examples)
+
+You can find examples for all supported features in the files in the
+[`tests/`](tests) directory. In case you are looking at these on GitHub, keep in
+mind that GitHub doesn't support some of the features that Vivify supports so
+some things may look off.
```
tuurep, line 41: Yeah I quite like this clarification because one could easily browse this on github file preview and think it's broken when it isn't for us :D

tests/markdown-extended.md
```diff
@@ -89,9 +89,9 @@ That is so funny! :joy:
 
 :cowboy_hat_face::nerd_face:
 
-## Highlight
+## Highlight/Mark
```
tuurep, line 92: Sure, don't mind things like these

Another one that has 2 different names is Definition list/Description list but I thought using only the first was okay in the context

tuurep (commented):
Yeah looks great!

tests/markdown-additional.md
```diff
@@ -28,3 +28,7 @@ digraph {
 While not a markdown syntax, this has a default style:
 
 Press <kbd>Ctrl</kbd> + <kbd>C</kbd> to copy, and <kbd>Ctrl</kbd> + <kbd>V</kbd> to paste!
+
+## Custom attributes
+
+This paragraph has a red background color.{style=background-color:red}
```
tuurep, line 34: Nice example, I was gonna try something like this but this pretty much covers it

tuurep:
> And of course the highlight color might be a topic for u as well ^^

Inspired by seeing a PR on the iamcco repo adding the same plugin:

https://github.com/iamcco/markdown-preview.nvim/pull/684

The color used in the screenshots there looks really good, and I think we could at least use the same one for light theme, and a similar for dark

image

What we currently have:

image

This is very crude and I don't like it :D

Do you agree?

However, in that PR they pretty much only added this same plugin, so I'll have to look at where the color is set

tuurep:
Fun fact/quirk:

Opening a new tab on a header anchor now adds a sort of 'selection' looking blue box around the element:

image

I'm, say, 65% sure this didn't happen before switching to using the `attr` plugin for this

This doesn't really matter to me but something I noticed as well..

tests/markdown-extended.md
jannis-baum: Yes I think it's fine. I just added it here because "mark" and "highlight" seem quite different ^^ definition vs. description list at least sounds like it could be the same thing

jannis-baum:
> This is very crude and I don't like it :D 

Yes, definitely! I think the PR that you referenced didn't commit everything they had to get to that screenshot (the code also has a typo and wouldn't compile). Maybe they just added some opacity to the plain yellow?

tuurep:
I also notice that in light theme it's the exact same color that Github has as background color in this:

image

So for light theme I think this is a good choice, it's `#fff8c5`

For the dark theme I'm not so sure. Do you have ideas?

image

Yellows that are dark enough look pretty much brown

Unless we flip the foreground:

image

Could be a different color too I guess, e.g.:

image

tuurep:
Github's choice in a dark theme:

image

Unexpected to me :smile:

jannis-baum:
How about `#ffffaf`? That's the color I use for search matches e.g. in Vim and ripgrep with my color scheme Jellyfish.

dark mode

I think it that exact same color also works surprisingly well in light mode

light mode

What do you think?

tuurep:
Hmm, it's pretty good

Since I imagine it's common to mark entire paragraphs and such with `====`, I wanted to see how it looks like in a larger segment:

mark-paragraph1


For light theme that actually looks great, here's the github mention color for comparison:

mark-paragraph2

This looks ok, too. But I might even slightly prefer yours.

For dark theme...

image

Ouch I think it stands out too much :smile: 

Do you agree however?

tuurep:
As a very personal take, if I was actually using the dark theme, I'd probably set it something close to this for myself:

image

in case that helps with decision making, but I want it to be primarily what you want :smile:

jannis-baum:
Ah nice, good idea about the paragraph. I also like the `ffffaf` on light and yes, dark might be slightly too strong.

I was also trying out some shades of purple and blue from my color scheme which I personally really like for the marks but I was wondering if this wouldn't be too far from the conventional yellow. What do you think?

tuurep:
Finding it pretty impossible to find a yellow bg for dark which looks good to me, yeah I think blue/purple is a good idea and conveys the meaning just fine as well

jannis-baum:
Actually, how about just adding some opacity? This would be the same yellow at 80% on both dark and light:

image

image

Or, if we want to use colors from the syntax highlighting for dark mode, I think these would be the best options:

1. `#d7afd7` image
2. `#d0d1fe` image

But I think I am slightly leaning towards keeping it simple and using the yellow at 80% for both. What's your take? After your next reply I will commit something and then merge this hahaha we're getting too deep into this :D

tuurep:
Ok! We can keep yellow for both, I would say the 0.8 opacity is an improvement on dark, but would prefer 1.0 on light (the original #ffffaf).

> we're getting too deep into this :D

Knowing myself I could tweak colors til the end of time so I do try to avoid bikeshedding but I think there was a need on this one :D

Hopefully the Github Alert colors will be easy on us

static/style.css
```diff
@@ -189,6 +184,12 @@ kbd {
     border-radius: 6px;
 }
 
+/* highlight/mark */
+mark {
+  background-color: rgba(255, 255, 175, 0.8);
+  color: #222;
```
tuurep, line 190: I couldn't think of this, (a little lighter fg) but I think this makes it subtly nicer looking in some way, nice

static/style.css
```diff
@@ -248,4 +249,8 @@ kbd {
         border: 1px solid #eff1f3;
         box-shadow: inset 0 -1px 0 #eff1f3;
     }
+
+    mark {
+        color: #1f2328;
```
tuurep, line 254: Can we set background-color: #ffffaf for light?

or could be written as rgba(..., 1.0) as well but this doesn't really matter

jannis-baum:
> Knowing myself I could tweak colors til the end of time so I do try to avoid bikeshedding but I think there was a need on this one :D

Haha yes, it's also nice that you put so much thought into this! Definitely helps the project

> Hopefully the Github Alert colors will be easy on us

PR #112, opened 2024-07-19
Version command

### LINKED ISSUE #111: Version command
`-v`/`--version`

jannis-baum:
Close #111

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,7 +36,7 @@ jobs:
       - name: build
         run: |
           yarn
-          make linux
+          VIV_VERSION="${{ github.ref_name }}" make linux
       - name: upload artifact
         uses: actions/upload-artifact@v4
         with:
@@ -53,7 +53,7 @@ jobs:
       - name: build
         run: |
           yarn
-          make macos
+          VIV_VERSION="${{ github.ref_name }}" make macos
       - name: upload artifact
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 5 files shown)


jannis-baum:
@tuurep sorry if I am being too annoying with changes/new features lately, it'll tone down a lot again soon🙈

Here I added a `-v`/`--version` to `viv` (by adding it to `vivify-server`) which I think is an important feature we have been missing.

For this to work I had to make some changes to `viv` that allow the things that `vivify-server` prints during startup to be output to the user through `viv`. This enables us to do all argument parsing (other than `--help`) in `vivify-server` and handle the version number there and also print an error if a requested file doesn't exist. Might also be useful for future additions.

For the releases the version number will be the same as the tag names, e.g. `v0.1.6`. For local builds in between it will be `{previous_tag}-{number_of_commits_since}-g{hash_of_head_commit}[-dirty]` with `-dirty` being added in case there are uncommitted changes, e.g. `v0.1.6-120-g05d074c` for the last commit I made here. Or, if `VIV_VERSION` is set during `make` it will be that. [^1]

[^1]: Not that relevant, but just for completeness: For non-release builds that the CI makes it will be `{PR-number}/merge`, e.g. `112/merge` here, and for the ones that happen on main simply `main`. Adding the full description like for local builds would require checking out the full repo with all the history which might be a bit unnecessary

Would you mind doing a quick check and seeing if the building and getting the version number works for you as well?

tuurep:
Yeah don't worry, it's been cool feeling productive on this project :D

The new flag works just fine:

```
$ viv --version
vivify-server v0.1.6-120-g05d074c
$ viv -v
vivify-server v0.1.6-120-g05d074c
```

tuurep:
```
$ VIV_VERSION="trolol-edition" make install
...
$ viv --version
vivify-server trolol-edition
```

PR #114, opened 2024-07-19
Fix release message body

### LINKED ISSUE #113: Fix release message body

jannis-baum:
Close #113


jannis-baum:
@tuurep Hey so a little update on the release: It almost worked haha. Only thing was that the tag's message didn't properly get inserted into the release notes, this should fix it now☺️

Problems were

1. We needed to use just `ref_name` and not the full `ref` with the command
2. The `content:subject` for whatever reason deletes all line breaks from the message so we have to get all `content` which ends in a PGP signature that we delete with `sed`

PR #118, opened 2024-07-22
GitHub alert block quotes

### LINKED ISSUE #89: GitHub alert block quotes
- [x] add markdown-it plugin for rendering
- [ ] figure out colors

---

OP @Tweekism in https://github.com/jannis-baum/vivify/issues/87#issuecomment-2226968861

> Right, so the reason I came looking for iamcco in the first place was to get these things working...
> 
> Screenshot from 2024-07-14 01-51-37
> 
> I have a young lady working in my office and she likes things to be pretty.
> 
> This is how it looks on GitHub:
> 
> Screenshot from 2024-07-14 01-52-44
> 
> Soooo.... add it to the list? 😅
            
Research from @tuurep 

> More information about those:
> 
> https://github.com/orgs/community/discussions/16925
> 
> Possible markdown-it plugin:
> 
> https://github.com/antfu/markdown-it-github-alerts
> 
> I think it goes more into the "additional" markdown but I would definitely be down to adding this

tuurep:
Close #89

### DIFF
--- a/src/parser/markdown.ts
+++ b/src/parser/markdown.ts
@@ -54,6 +54,9 @@ mdit.use(attributes);
 
 // MARK: untyped plugins done
 
+import githubAlerts from 'markdown-it-github-alerts';
+mdit.use(githubAlerts);
+
 // anchor has to be added after attribute plugin for ids to work
 mdit.use(anchor, {
     permalink: anchor.permalink.ariaHidden({
--- a/src/parser/markdown.ts
+++ b/src/parser/markdown.ts
@@ -2,6 +2,7 @@ import MarkdownIt from 'markdown-it';
 import anchor from 'markdown-it-anchor';
 import highlight from './highlight.js';
 import graphviz from './dot.js';
+import githubAlerts from 'markdown-it-github-alerts';
 import config from './config.js';
 import { Renderer } from './parser.js';
 
@@ -15,47 +16,44 @@ const mdit = new MarkdownIt({
 // ts-expect-error for blocks so this is ugly:
 // https://github.com/Microsoft/TypeScript/issues/19573
 
-/* @ts-expect-error: markdown-it modules aren't typed */
+/* @ts-expect-error: module not typed */
 import { full as emoji } from 'markdown-it-emoji';
 mdit.use(emoji);
-/* @ts-expect-error: markdown-it modules aren't typed */
+/* @ts-expect-error: module not typed */
 import taskLists from 'markdown-it-task-lists';
 mdit.use(taskLists);
-/* @ts-expect-error: markdown-it modules aren't typed */
+/* @ts-expect-error: module not typed */
 import footNote from 'markdown-it-footnote';
 mdit.use(footNote);
-/* @ts-expect-error: markdown-it modules aren't typed */
+/* @ts-expect-error: most module not typed */
 import lineNumbers from 'markdown-it-inject-linenumbers';
 mdit.use(lineNumbers);
-/* @ts-expect-error: markdown-it modules aren't typed */
+/* @ts-expect-error: module not typed */
 import texMath from 'markdown-it-texmath';
-/* @ts-expect-error: markdown-it modules aren't typed */
+/* @ts-expect-error: module not typed */
 import katex from 'katex';
 mdit.use(texMath, {
     engine: katex,
     delimiters: 'dollars',
     katexOptions: config.katexOptions,
 });
-/* @ts-expect-error: markdown-it modules aren't typed */
+/* @ts-expect-error: module not typed */
 import defList from 'markdown-it-deflist';
 mdit.use(defList);
-/* @ts-expect-error: markdown-it modules aren't typed */
+/* @ts-expect-error: module not typed */
 import sub from 'markdown-it-sub';
 mdit.use(sub);
-/* @ts-expect-error: markdown-it modules aren't typed */
+/* @ts-expect-error: module not typed */
 import sup from 'markdown-it-sup';
 mdit.use(sup);
-/* @ts-expect-error: markdown-it modules aren't typed */
+/* @ts-expect-error: module not typed */
 import mark from 'markdown-it-mark';
 mdit.use(mark);
-/* @ts-expect-error: markdown-it modules aren't typed */
+/* @ts-expect-error: module not typed */
 import attributes from 'markdown-it-attrs';
 mdit.use(attributes);
 
 // MARK: untyped plugins done
-
-import githubAlerts from 'markdown-it-github-alerts';
-mdit.use(githubAlerts);
 
 // anchor has to be added after attribute plugin for ids to work
 mdit.use(anchor, {
@@ -64,6 +62,7 @@ mdit.use(anchor, {
     }),
 });
 mdit.use(graphviz);
+mdit.use(githubAlerts);
 
 const renderMarkdown: Renderer = (content: string) => {
     return mdit.render(content);
--- a/static/style.css
+++ b/static/style.css
@@ -2,8 +2,8 @@
     --color-alert-note: #a5d5fe;
     --color-alert-tip: #b4fa72;
     --color-alert-important: #ff8ffd;
-    --color-alert-warning: #fefdc2;
-    --color-alert-caution: #ff8272;
+    --color-alert-warning: #ffaf00;
+    --color-alert-caution: #ff5f5f;
 }
 
 html {
--- a/static/style.css
+++ b/static/style.css
@@ -1,3 +1,11 @@
+:root {
+    --color-alert-note: #a5d5fe;
+    --color-alert-tip: #b4fa72;
+    --color-alert-important: #ff8ffd;
+    --color-alert-warning: #fefdc2;
+    --color-alert-caution: #ff8272;
+}
+
 html {
     background-color: black;
     color: #aaa;
@@ -122,7 +130,6 @@ section.footnotes {
 
 /* --------------------------------------------------------------------------
  * DEFIITION LISTS ---------------------------------------------------------- */
-
 dl {
     display: flex;
     flex-flow: row wrap;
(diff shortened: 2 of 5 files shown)


src/parser/markdown.ts
```diff
@@ -54,6 +54,9 @@ mdit.use(attributes);
 
 // MARK: untyped plugins done
 
+import githubAlerts from 'markdown-it-github-alerts';
+mdit.use(githubAlerts);
```
tuurep, line 58: @jannis-baum

All other plugins have this `/* @ts-expect-error: markdown-it modules aren't typed */` before the imports, but this one would error out of build with that.

Is this really the first "typed" plugin?

jannis-baum: The one you are using is one of the few that are actually typed ^^ You can tell by the blue "TS" logo next to the heading: https://www.npmjs.com/package/markdown-it-github-alerts

Maybe this means we should adjust the comments to say "most ...". The graphviz and anchor plugins in the bottom of that file are also typed.

tuurep: Oh yes, the "markdown-it modules aren't typed" is just a comment and not a string match to the error? :smile:

tuurep: Ok I can move GHAlert import to the top like done with the other ones and adjust the comments as well

jannis-baum: > Oh yes, the "markdown-it modules aren't typed" is just a comment and not a string match to an error message? 😄

It looks like a comment but it's actually a directive for the TS compiler. `@ts-expect-error` tells it to ignore the error that will happen for lack of types in the next line and just keep the code as it is. The part after that is a mandatory description of the error that I wrote because it will complain and still fail if there isn't one ^^

tuurep: Lol what a pain

I think since you have the full info at the top comment, they can just say "module not typed" or w/e?

tuurep:
Ok @Tweekism @jannis-baum this is ready for feedback

I basically took what's here:

https://github.com/antfu/markdown-it-github-alerts/blob/main/styles/github-base.css

But fit it into our `style.css`, maybe removing some redundant lines

Screenshots:

ghalert1
ghalert2

My concerns:
* Would there be a way to plug the styles from the plugin without essentially copy pasting CSS?
* What do you think about those ANSI colors on dark? To me looks a bit hit and miss.. Some are ok, but some too light, notably yellow.

static/style.css
```diff
@@ -1,3 +1,11 @@
+:root {
+    --color-alert-note: #a5d5fe;
+    --color-alert-tip: #b4fa72;
+    --color-alert-important: #ff8ffd;
+    --color-alert-warning: #fefdc2;
+    --color-alert-caution: #ff8272;
+}
```
tuurep, line 7: How do we feel about color variables on `:root` like this?

Note how it makes setting the colors for light theme *far* less verbose than it would be without these.

jannis-baum: Yes, this is true. I like it a lot. But I would prefer if we could refactor and do this for everything then🙈 Would you mind looking into that? It would make the CSS much easier to read and use

Since we're already talking about refactoring CSS: We could also consider to start using something like https://lesscss.org or some other compiler for the CSS to make it even easier to work with. What do you think? If you like this or another one I can add it to the dev and build infrastructures

jannis-baum:
Hey thanks! I will look into the code later or tomorrow :) 

> Would there be a way to plug the styles from the plugin without essentially copy pasting CSS? Are they on the repo just for reference or supposed to be able to be used directly?

Not sure. On giving their docs one short look I couldn't find instructions. Might also be that our set up is a bit too exotic for this to work. One cheeky thing you could try however is making a symlink in the static directory to their styles at `node_modules/....css` and then adding that symlink to be included in the as styles in our `viewer.ts` ^^ Might just work. If it works on dev but not for the build we can fix it. If it doesn't work on dev it would get a bit ugly I think

Tweekism:
> * What do you think about those ANSI colors on dark? To me looks a bit hit and miss.. Some are ok, but some too light, notably yellow.
>       
>   * They are kinda ok though :)

These colors are workable, I missed the context for why ANSI colors. Is the intention to use the same colors in the terminal in vim?

static/style.css
Tweekism: Is it a lot of work to change all the css to this style?

Would it be worth letting this merge and the re-unifying the rest with a seperate PR?

I don't know enough about css compliers. Sometimes less is more, but maybe the benefits are enough to justify the additional complexity. Would it make it more confusing for users to then implement their custom css files?

jannis-baum: > Is it a lot of work to change all the css to this style?

Somewhat much but I think it'd be worth it for more unified colors & easier maintenance and adding new features

> Would it be worth letting this merge and the re-unifying the rest with a seperate PR?

Yes, definitely

> I don't know enough about css compliers. Sometimes less is more, but maybe the benefits are enough to justify the additional complexity. Would it make it more confusing for users to then implement their custom css files?

It would make the build infrastructure slightly more complex but the code would get a lot more readable. User-CSS would not be affected at all

tuurep: >> Would it be worth letting this merge and the re-unifying the rest with a seperate PR?
>
> Yes, definitely

+1 from me, I don't think there's necessarily a problem with having just these in `:root`

tuurep: > But I would prefer if we could refactor and do this for everything then🙈 Would you mind looking into that?

Yeah lets start an issue

jannis-baum: > Yeah lets start an issue

Nice, can you open it? Let's debate there if we also want to introduce a CSS compiler in the same context :)

tuurep:
> > What do you think about those ANSI colors on dark? To me looks a bit hit and miss.. Some are ok, but some too light, notably yellow.
> 
> I honestly don't like them that much to be honest🙈

That's a good answer :smile: What kind would work best?

My thought is that they could be like the light theme colors but adapted to dark, then adjust and whatever

Wdyt about that or did you have something else in mind?

I can make an initial suggestion if needed

jannis-baum:
> That's a good answer 😄 What kind would work best?
> 
> My thought is that they could be like the light theme colors but adapted to dark, then adjust and whatever
> 
> Wdyt about that or did you have something else in mind?
> 
> I can make an initial suggestion if needed

I honestly don't have any idea ^^ Would be glad about a suggestion! I did see however that the plugin also has a dark theme CSS (https://github.com/antfu/markdown-it-github-alerts/tree/main/styles) so if you want to go the route of adjusting the light colors to dark that might already be done there :)

tuurep:
> I did see however that the plugin also has a dark theme CSS (https://github.com/antfu/markdown-it-github-alerts/tree/main/styles) so if you want to go the route of adjusting the light colors to dark that might already be done there :)

Yeah I can try this as an initial step

The dark theme is "GitHub Dark" and notable that it has quite a different background color than yours

I could also look at GitHub High Contrast alert colors

jannis-baum:
> But I'm kinda interested in what the function of the CSS in the repo is. Whether that's some markdown-it plugin convention and there's more info about that. I'll see.

I think it's some bigger thing than just `markdown-it`; I have definitely seen similar ways of importing CSS in NextJS and React so there is probably some big widely used library that does this. I don't know how applicable it is to our use-case of serving the file statically with express though. If we want to do that we might have to change quite a bit of stuff and start using JSX with some other dependency.

jannis-baum:
> The dark theme is "GitHub Dark" and notable that it has quite a different background color than yours

Good point!

> I could also look at GitHub Dark High Contrast alert colors

Yes true. Might just be that those colors "pop" a bit too much but maybe it's a good starting point anyways.

tuurep:
:warning: I'll go deep into GitHub colors

The markdown-it plugin repo CSS has slight differences in colors to what they actually are on GitHub as of this moment, but it does look like they were taken from the GitHub variables directly, ~8 months ago:

https://github.com/antfu/markdown-it-github-alerts/commit/595f57831879f6058bf5df430357747b21cdfd74

(due to the same variable names that you see when you inspect GH)

Speculation: I guess that GH adjusts these colors over time, and CSS such like in the plugin doesn't keep up

So I would take the colors straight from inspecting GH. But lets look at screenshots for:

1. Plugin CSS
2. Straight from GitHub CSS
    * For dark: GH Dark and GH Dark High Contrast

# Light

* Only the red and yellow (warning/caution) are different

## Plugin repo CSS

image

## GitHub up-to-date CSS

image

# Dark

* All colors are different

## Plugin repo CSS

image

## GitHub up-to-date Dark CSS

image

## GitHub up-to-date Dark High Contrast CSS

image

tuurep:
So in my opinion any of the last 3 (note: I added a "low contrast") could work for dark, and all look pretty good

I would avoid the outdated plugin CSS

Thoughts @jannis-baum ?

tuurep:
Default:

`#1f6feb`  
`#238636`  
`#8957e5`  
`#9e6a03`  
`#da3633`

High contrast:

`#409eff`  
`#09b43a`  
`#b87fff`  
`#e09b13`  
`#ff6a69`

Dimmed:

`#316dca`  
`#347d39`  
`#8256d0`  
`#966600`  
`#c93c37`

Tweekism:
Ok yeah the up to date dimmed does look a lot nicer.

tuurep:
Yeah it could be a fit!

To be perfectly clear the Dimmed one is not something I created, but it's a theme on GitHub:

image

static/style.css
tuurep: Alright: https://github.com/jannis-baum/Vivify/issues/120

jannis-baum:
Hey, sorry I've been taking so long to reply to this, I've been back and forth on the colors because I am not that happy with any of them. They all seem a bit dull to me and the warning color is very brown.

I just checked through Jellyfish again and realized I have colors for compiler notes, warnings and errors that might be quite suitable. So if we use those three plus the ANSI colors for "tip" and "important" it would look like this:

image

```css
:root {
    --color-alert-note: #afafff;
    --color-alert-tip: #b4fa72;
    --color-alert-important: #ff8ffd;
    --color-alert-warning: #ffaf00;
    --color-alert-caution: #ff5f5f;
}
```

What do you think?

Tweekism:
I'm a little triggered that **Note** has shifted from blue to more of a purple 😂

And it might be a little too similar to **Important**?

tuurep:
I was just about to comment too

It's definitely an improvement from the first

(this)
alert1

I think it's pretty fine tbh and I'd approve, but if we need to adjust then lets think about it

Safe to say lets not use the GH colors directly, if they look out of place in the theme

jannis-baum:
Soooo how about falling back to the ANSI color for `Note` as well?

image

```css
:root {
    --color-alert-note: #a5d5fe;
    --color-alert-tip: #b4fa72;
    --color-alert-important: #ff8ffd;
    --color-alert-warning: #ffaf00;
    --color-alert-caution: #ff5f5f;
}
```

Tweekism:
I like it, I like that our pop a little more. I'd be happy with either honestly.

I know no one was asking, but here is an up close comparison with the git hub ones.

image

Tweekism:
Its easy to tweak a single colour in custom css, but it is important to have sensible defaults

tuurep:
> it is important to have sensible defaults

Yeah I agree with this very much :)

I have a little feeling that the blue is just a bit too light, similarly to the ANSI yellow but not nearly as severe.

But the more I've stared it and tried to come up with an improvement, the more I think it's kinda OK enough.

jannis-baum:
Hahaha okay so let's just merge this? In case we end up disliking it too much we can tweak it when we refactor the CSS in #120. That one will probably be a huge color discussion anyways😂

jannis-baum:
Oh but before we merge, @tuurep do you want to commit the latest version we agreed on now? I.e. ANSI plus the warning and error colors.

tuurep:
Alright, done!

PR #143, opened 2024-07-29
Document API & architecture

### LINKED ISSUE #119: Document API & architecture
To make it easier for people to contribute & integrate Vivify into other tools

jannis-baum:
Close #119

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,7 +1,7 @@
 # Vivify's API
 
 In addition to the command line interface `viv`, Vivify has HTTP endpoints that
-can be used to integrate it into other applications, such as making editor
+are be used to integrate it into other applications, such as making editor
 plugins for live previews.
 
 ## Path encoding
@@ -10,15 +10,15 @@ Vivify's endpoints are often accessed with the pattern `/<endpoint>/<path>`
 where path is a
 [percent-encoded](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding)
 path in the file system. For paths that follow UNIX naming conventions, this
-means that endpoints can be accessed directly at the literal paths. For example,
-accessing `<endpoint>` for a file at `/Users/me/my-file.md` can be done through
-the URL
+means that endpoints are be accessed directly at the literal paths. For example,
+accessing `<endpoint>` for a file at `/Users/me/my-file.md` is done through the
+URL.
 
 ```txt
 /<endpoint>/Users/me/my-file.md
 ```
 
-For paths that have special characters such as spaces, we have to rely on the
+For paths that have special characters such as spaces, use
 [percent-encoding](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding).
 E.g. to access `<endpoint>` for a file at `Users/me/my file.md`, it will be the
 following.
@@ -68,5 +68,4 @@ The request can include a body with the following optional fields.
 
 Sending a `DELETE` request to `/viewer` will fully reload all connected clients
 at any paths and clear all cached updated live content. Sending a `DELETE`
-request to `/viewer/<path>` (as defined in [Path encoding](#path-encoding)) will
-do the same for all clients at the given path.
+request to `/viewer/<path>` will do the same for all clients at the given path.


jannis-baum:
Ah, one more thing: This is for the state after #117 is merged

Tweekism:
Cool, lemme have a look.

Tweekism:
The API stuff looks really good! I think that is a big help to anyone wanting to use the API. There is a couple things I could point at if you like, I'll do that as a review thing, but you could publish it as is honestly.

I'll start working on the architecture one as soon as I can. I'll go through it slowly and make sure I understand it as I add to it.

docs/api.md
```diff
@@ -0,0 +1,72 @@
+# Vivify's API
+
+In addition to the command line interface `viv`, Vivify has HTTP endpoints that
```
Tweekism, line 3: Now, this might be silly, so you can choose to adopt this or not, but I would propose that the project name always be wrapped in bold tags, so that **Vivify** would always appear as bold, as like a branding thing.

```suggestion
In addition to the command line interface `viv`, **Vivify** has HTTP endpoints that
```

docs/api.md
```diff
@@ -0,0 +1,72 @@
...
+can be used to integrate it into other applications, such as making editor
+plugins for live previews.
+
+## Path encoding
+
+Vivify's endpoints are often accessed with the pattern `/<endpoint>/<path>`
```
Tweekism, line 9: Would it be fair to say they are always accessed with that pattern? I would consider just dropping the word often.

docs/api.md
```diff
@@ -0,0 +1,72 @@
...
+where path is a
+[percent-encoded](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding)
+path in the file system. For paths that follow UNIX naming conventions, this
+means that endpoints can be accessed directly at the literal paths. For example,
+accessing `<endpoint>` for a file at `/Users/me/my-file.md` can be done through
+the URL
```
Tweekism, line 15: This one is missing its ending period
```suggestion
the URL.
```

docs/api.md
```diff
@@ -0,0 +1,72 @@
...
+
+```txt
+/<endpoint>/Users/me/my-file.md
+```
+
+For paths that have special characters such as spaces, we have to rely on the
+[percent-encoding](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding).
```
Tweekism, line 22: Suggestion, use authoritative tone
```suggestion
For paths that have special characters such as spaces, use
[percent-encoding](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding).
```

docs/api.md
```diff
@@ -0,0 +1,72 @@
...
+                        // corresponding to the given line in the source file.
+  "reload": boolean     // Live update the viewer based on the current content
+                        // on disk. Note that this overrides "content".
+}
+```
+
+Sending a `DELETE` request to `/viewer` will fully reload all connected clients
+at any paths and clear all cached updated live content. Sending a `DELETE`
+request to `/viewer/<path>` (as defined in [Path encoding](#path-encoding)) will
+do the same for all clients at the given path.
```
Tweekism, line 72: I would drop the "as defined in..." after the first use in each section (ie: one use in the health section, one use in the view section at maximum)

```suggestion
request to `/viewer/<path>` will do the same for all clients at the given path.
```

docs/architecture.md
Tweekism, line 1: Since this is a rough draft, probably makes more sense if I edit it offline using my local text editor and this awesome markdown viewer I found called **Vivify** 🤣

Tweekism (commented):
I made some notes. 

I'll take the bigger architecture.md edit offline.

docs/api.md
jannis-baum: Hm, I see the point but I feel like that might be a bit cumbersome to do because we would have to do it in all other files as well and will probably forget some, especially as the documentation evolves. I would say the capitalization takes care of this to a good enough level since most other words are lower case in English. What do you think?

docs/api.md
jannis-baum: We do have `GET /health` and `DELETE /viewer` at the root of the endpoints with no paths.

We could instead say something like

> **Current wording**
>
> > Vivify's endpoints are often accessed with the pattern `/<endpoint>/<path>` where path is a [percent-encoded](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding) path in the file system.
>
> **Suggestion**
>
> > When accessing an endpoint corresponding to a path in the file system, use `/<endpoint>/<path>` where `<path>` is [percent-encoded](https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding).

Is that better? ^^

docs/architecture.md
jannis-baum: Yes, sounds good! That's the idea that you write the whole thing haha, the stuff I wrote into the document is only for you, not for `main` :D

docs/api.md
Tweekism: Yeah that works fine. That's certainly easier.

docs/architecture.md
Tweekism: Rgr. Do you plan to merge api.md and leave architecture.md out? I've already pulled it down so I can keep working on it.

jannis-baum: Ah, I don't mind. I was thinking to leave this PR laying around until you are done with the architecture document, I don't think we are in a big rush here. Or what do you think?

Tweekism: Fine by me, I am... thinking about putting in a job app this week, and then its my weekend for kid wrangling duties, so I might not get to it soon is all.

PR #150, opened 2024-08-01
Parse empty files

### LINKED ISSUE #149: Parse empty files

jannis-baum:
Close #149

### DIFF
--- a/src/parser/parser.ts
+++ b/src/parser/parser.ts
@@ -34,7 +34,7 @@ function textRenderer(
 }
 
 export const shouldRender = (mime: string): boolean =>
-    mime.startsWith('text/') || mime === 'application/json';
+    mime.startsWith('text/') || mime === 'application/json' || mime === 'inode/x-empty';
 
 export function renderTextFile(content: string, path: string): string {
     const fileEnding = path?.split('.')?.at(-1);


jannis-baum:
One caveat here is of course that you have to save the file before you can open Vivify, on a new buffer that is not saved yet this won't work because Vivify is organized in parallel to the file system

tuurep:
Yes, exactly what I was after! Thanks!

Also tested with an extensionless file, and opens this:

image

Whereas before it would have asked me if I want to save file, because of this firefox setting I think:

image

And showing the empty text file is much better in my opinion

tuurep (approved):

jannis-baum:
Nice! Happy this was so easy for once hahahaha

> Whereas before it would have asked me if I want to save file, because of this firefox setting I think:

Yeah files with `shouldRender(...) == false` will be sent as data which is necessary e.g. to render local images in Markdown files. And browsers will do the download when they get data.

Tweekism:
On Windows, this will be `application/x-empty`

Mental note for me for later...

tuurep:
Btw this doesn't behave so well when you delete all contents in a file while previewing,

* rather than empty buffer, shows last non-empty contents
* if you save it as empty and re-view, still shows the last non-empty contents

But I'm not sure how much is related to:

> Vivify is organized in parallel to the file system

Is it changing the `file` -type on the fly, if you know what I mean?

jannis-baum:
@tuurep Ah okay, this is probably because the empty string resolves to "false" so it doesn't update it and then it still has the live content saved when you reopen it. I'll fix that later :)

tuurep:
Nice, sounds like an easy problem :smile:

Issue #101, opened 2024-07-16
Set Notebook ANSI colors


jannis-baum:
continued discussion from #89

- [ ] ANSI colors that are parsed by [ansi_up](https://www.npmjs.com/package/ansi_up)
- [ ] error outputs, `.ipynb-output-stream-stderr` and `.ipynb-output-error` in CSS

jannis-baum:
@tuurep feel free to add to the list above if there's anything else and assign yourself to the issue if you want to work on this :)

jannis-baum:
@tuurep

> > There we also have to render ANSI and the current default colors are very ugly
> 
> I've only taken a brief glance and wasn't overly horrible on Light Theme, but does it use the same colors for both Light and Dark?

I created a branch for this issue and added some more code to the notebook to look at the main colors. As you suspected, especially in dark mode it looks bad right now:

<img width="1268" alt="image" src="https://github.com/user-attachments/assets/b14ea1e6-6d15-489e-96c5-b8d7cff4fd96">

tuurep:
Thanks, it's good info! I'd be interested looking into this but only after a few other things first.

As for the borderline-unreadable white-on-yellow error text, I remember seeing this issue in the past:

https://github.com/ipython/ipython/issues/13446

If that ends up a problem, will be useful to look there first.

jannis-baum:
Interesting, good research on that!

> I'd be interested looking into this but only after a few other things first.

Nice! What's your agenda for the next things you want to do on Vivify? Just to keep an overview :)

tuurep:
I'll first review your config PR, then add GH alerts plugin

After that I think I'll take a look at this issue to know whether it's simple or complicated

And my two issues on the vim plugin repo are constantly on my mind :D

I've been using a branch for myself which refreshes on `TextChanged` and `TextChangedI` and it has worked well enough but would be great to do it properly

jannis-baum:
> I'll first review your config PR, then add GH alerts plugin
> 
> After that I think I'll take a look at this issue to know whether it's simple or complicated

Thank you! Sounds good!
 
> And my two issues on the vim plugin repo are constantly on my mind :D

[...]

PR #58, opened 2024-01-03
Table style, horizontal rule style, linkification

tuurep:
Issues: #56 #57

I think tables look great now:

image

image

I don't like the way Github styles horizontal rule too much, but I didn't have a better suggestion so copied the Github one for a default:

image

image

Do you think these are good to go?

PR #151, opened 2024-08-01
CSS refactor

### LINKED ISSUE #120: Consider CSS refactor
- [x] split up the current styles.css into a dedicated markdown.css file and another one that has everything else that's more general
- [x] extract all the colors into variables and remove the media query for light mode there
- [x] add two files colors-dark.css and colors-light.css where we do have media queries and just define all the color variables

### Original description

Followup from discussion https://github.com/jannis-baum/Vivify/pull/118#discussion_r1687058052

Regarding color variables like this:

```
:root {
    --color-alert-note: #a5d5fe;
    --color-alert-tip: #b4fa72;
    --color-alert-important: #ff8ffd;
    --color-alert-warning: #fefdc2;
    --color-alert-caution: #ff8272;
}
```

> I would prefer if we could refactor and do this for everything then🙈 Would you mind looking into that? It would make the CSS much easier to read and use

Ideas about adding more tooling for CSS:

> Since we're already talking about refactoring CSS: We could also consider to start using something like https://lesscss.org/ or some other compiler for the CSS to make it even easier to work with. What do you think? If you like this or another one I can add it to the dev and build infrastructures

jannis-baum:
Close #120


jannis-baum:
Hey @tuurep so I did this now, I think it made the CSS a lot less cluttered and will hopefully make it easier to maintain & extend in the future :)

I concentrated as hard as I could while doing this, I hope I didn't mess anything up ^^ If you want to check this I would recommend looking at each of the two commits whose messages are `refactor(#120): extract ... colors` individually, this way it should be easiest to figure out if I made any mistakes.

Also let me know if you're happy with this or if you would refactor anything else :) I did a couple of other small things, see commit messages

tuurep:
Hey, looks great. Even surprises me how few variables we end up needing.

Tested and looks as expected on light theme. I only have naming-related concerns, I'll put a review.

static/colors.css
```diff
@@ -0,0 +1,96 @@
...
+        --bg-code: #afb8c133;
+        --bg-mark: rgba(255, 255, 175, 1);
+
+        --text-primary: #1f2328;
+        --text-secondary: #656d76;
+        --text-link: #0969da;
+        --text-mark: var(--text-primary);
+
+        --border-light: #d8dee4;
+        --border-dark: #d0d7de;
```
tuurep, line 63: Naming works for dark theme, but for light theme these are essentially vice-versa: light is dark and dark is light

Some alternative suggestions:

--border / --border-dimmed
--border / --border-faint
--border-regular / ...

static/colors.css
```diff
@@ -0,0 +1,96 @@
+/* --------------------------------------------------------------------------
+ * DARK MODE ---------------------------------------------------------------- */
+
+:root {
+    --bg: black;
+    --bg-highlight: #161616;
```
tuurep, line 6: Why's it called `highlight`? :smile: It took me a while to figure out what this color is for. My expectation would be `codeblock`

Or perhaps `pre` but that's only clear if you know what the tag's called.

What do you think, could this be `--bg-codeblock`?

tuurep (changes requested):

tuurep:
That's all!

Drawback of this is that we will have to come up with new variables and sort of maintain a naming scheme for these, but looks like it's worth it IMO.

static/colors.css
jannis-baum: Haha funny, I actually did this by just giving the **light** mode media query part of the original `styles.css` to GPT4 and having it find names for the colors. I modified some names so they are more structured, but `border-light` was one I left the way GPT named it. Then later when I did the dark mode colors I checked if `border-light` would "also" be the lighter one in dark mode, surprisingly saw that it was, and assumed everything would fit :D

static/colors.css
jannis-baum: I was trying to find something that encapsulates

1. the `pre`/code-blocks,
2. the alternating highlighted colors of tables, and
3. the `kbd`-tag background.

That's why I think `codeblock` or `pre` would be too specific.

We could do `bg-secondary` and rename the current `bg` to `bg-primary`. That would match the `text-` colors nicely and I think it's also at least somewhat conventional to have a "secondary background color" defined that should work for stuff that's on top of the "primary background color". What do you think?

static/colors.css
jannis-baum: How about we take the naming from how font weights are named?

https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#common_weight_name_mapping

I would suggest

```txt
border-light → border-light
border-dark  → border-medium
border-kbd   → border-bold
```

We could also name `border-light` as `border-regular` because that is the most "regular" border color, but I'm not sure how clear it is to people that "medium" is stronger than "regular". Or another option might be sticking more to the applicable [CSS values](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#values):

```txt
border-light → border-normal
border-dark  → border-bold
border-kbd   → border-bolder
```

OR taking something from [Tailwind's naming](https://tailwindcss.com/docs/font-weight):

```txt
border-light → border-normal
border-dark  → border-semibold
border-kbd   → border-bold
```

Maybe this is the most sensible solution? I think if anyone knows that they are doing with naming style-related things it's Tailwind ^^

static/colors.css
Tweekism: GitHub refers to these as "muted"

image

tuurep: Ah that's right, didn't realize how generic the color is.

> We could do `bg-secondary` and rename the current `bg` to `bg-primary`. That would match the `text-` colors nicely and I think it's also at least somewhat conventional to have a "secondary background color" defined that should work for stuff that's on top of the "primary background color". What do you think?

I like this idea the most :+1:

static/colors.css
tuurep: Hmmmm it's complicated

One thing I notice that on light theme the kbd border certainly doesn't correlate with `bold`, rather it's the faintest border I've ever seen in my life:

image

I think it's tricky because I'd say all border colors we have are very faint but slightly different.

Other than that, good ideas. Just I'm not sure which one should be considered the `normal` one, is it the <h1> bottom border color?

jannis-baum: We could also consider putting the whole border definition into the variable. E.g. make the value `0.5px solid #...`. This way the naming by "bold" and so on would be easier and the border width becomes part of the themes

tuurep: Oh, is the width what you were thinking about with `bold` I had the darkness/lightness of color in my mind still :smile:

jannis-baum: No haha, I was thinking about both. So maybe it's good to include both. Also because a different theme with a different color might need a different width to look good

tuurep: Yeah, I worry about a slippery slope towards a mess though if we include variables other than colors (at least right now)?

image

Here are the border colors side-by-side btw,

My perception is like (left-to-right):

* normal
* slightly more pronounced
* a special case for kbd

jannis-baum: Not really any solution, just more overview:

Colors for both themes (same order, currently "light" → "dark" → "kbd"):

image

How it looks in practice

image
image

Markdown to reproduce:

```markdown
## header

<kbd>kbd</kbd>

> block <kbd>kbd</kbd>

---

| table[^1] |
| :-- |
| <kbd>kbd</kbd> |
| <kbd>kbd</kbd> |
| |
| |

[^1]: footnote
```

jannis-baum: As a follow-up on my idea to include the whole border style, here are the distince styles we currently have[^1]:

```css
h1, h2, section.footnotes, dd /*definition lists*/ {
  border: 0.2px solid var(--border-light);
}

blockquote {
  border: 0.25em solid var(--border-dark);
}

hr /*horizontal rule*/ {
  border: 24px solid var(--border-dark);
}

th, td /*tables*/ {
  border: 1px solid var(--border-light);
}

kbd {
  border: 1px solid var(--border-kbd);
}
```

Looking at this I think we could unify `blockquote` and `hr` to use the same style, and also the tables and `kbd`. This would give us more consistent styling and maybe make it easy to name all 3 styles.

[^1]: note that I wrote `border` everywhere even though it might be `border-left` or another side for some examples but this wouldn't break from using the variables as I suggested

tuurep: Ok, it's tripping me up big time that I chose a color back in #58 opposite to how I'd expect it to be as a light-theme counterpart, but it still looks appropriate to me and maybe it was deliberate.

\<kbd> doesn't look good in alternate table cell but not too unexpected and I suppose GitHub has the same issue?

jannis-baum: Okay so including the whole border definition in 1 variable doesn't work well because we need *only* the color to make the shadow for the `kbd`-element. So back to just colors. I still liked the idea of getting rid of one of the 3 colors though and made a suggestion now in the latest commit and as described below.

**Suggestion** We keep only 2 border colors: `border-regular` for slim borders less than 1px and `border-bold` for borders from 1px. The screenshots below show the before/after of this change, left is before, right is after.

image
image

I particularly like (1) that the `kbd` element has much better visibility for light mode now, and (2) the table borders aren't as overly contrasty anymore for dark mode. Additionally, we would of course reduce complexity of the CSS/colors this way since we get rid of the specialized color that is only used for the `kbd` element.

What do you think?

tuurep: I'm kinda wary about changing some of the border colors from what they were on GitHub when they look just fine as is, but also have a bit of a dilemma because the border you suggest for \<kbd> looks much better, no doubt...

Maybe I can word what I'm thinking like this:

* We don't have to follow GitHub's style to a tee, when their style is not good (their \<kbd> border is so faint you can hardly see it)
* When there's nothing wrong with the GitHub style (e.g. light theme table border *imo*), not sure we should change that :thinking: 

I still feel like there should be 2 border colors that can be applied freely to anything without any regard to border thicknesses, for example on GH these same thickness borders are like (with their var names):

image

However it's further complicated by that I agree with you that the fainter table border looks better on dark theme. How would that fit into this then...

I have to really properly investigate this but if this gives you any thoughts and if we have a disagreement, let me know :smile:

tuurep: But one thing I'm sure: removal of border-kbd variable in favor of something that looks better on it is an absolute +1 from me

jannis-baum: Hm, I don't really have the goal of following GitHub's style at all to be honest haha, our dark theme is also completely different anyways. What's important for me is to have something that (1) looks good with everything being well-visible, and (2) is as simple as possible because (other than GitHub) we are not a huge team working full time on this and are (at least talking for myself) not professional designers :D

Another argument for favoring simplicity is that the state of this branch makes it very easy to create a Vivify color theme (i.e. a CSS file that just defines color variables) and creating one for GitHub is insanely hard and complicated because of the huge mountain of hard-to-understand colors that they define. I get taking GitHub as a kind of "gold standard" but I think its file-viewing goals diverge from those of Vivify in some important aspects (such as customizability).

And, finally, if our goal is to follow GitHub's styles, that will also mean we have to regularly check *everything* we take from their styles to make sure we are still up-to-date.

---

So, long story short (sorry if that was too much of a rant ^^) I would merge this as it is, but if you have another idea that takes my arguments into account, please feel free to suggest something different!

jannis-baum: > I still feel like there should be 2 border colors that can be applied freely to anything without any regard to border thicknesses, for example on GH these same thickness borders are like (with their var names):

We can also still do this, just have to come up with names. GitHub's "-muted" might be an option. Like I said, please make suggestions if you have ideas :) I'm just starting to run out of ideas for this haha and would like to merge soon to do #101 based on this

tuurep: Yeah no worries, I think we can merge today

> And, finally, if our goal is to follow GitHub's styles, that will also mean we have to regularly check *everything* we take from their styles to make sure we are still up-to-date.

This is a big point that was inevitably going to catch up to us, in fact if I'm not wrong some of the colors may have already been slightly adjusted... and I was wondering whether we aim to follow or what, and yeah definitely lets not try to keep up with that.

I was confused above with borders `<1px` because they end up being `1px` on my machine so I thought the changes here meant that h1 bottom border and table border would be the same, but it's not the case. So actually light mode is just fine!

How do you feel about this discrepancy:

* On light mode, regular is muted (less contrast)
* On dark mode however, bold is the muted one

^ a decision still resulting from my choice of color in #43 for quote border

So if I fix this logical inconsistency, for example in this way:

* border-regular: `#444444` --> `#303030`
* border-bold: `#353535`

^ (keeping the border color we currently like and readjusting regular around that)

This is how it would look like:

image

^ With that, we'd have the same definition for border-regular and border-bold regardless of light/dark

Do you think we wanna make this change?

static/colors.css
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/* --------------------------------------------------------------------------
+ * LIGHT MODE --------------------------------------------------------------- */
+
+@media (prefers-color-scheme: light) {
+    :root {
+        --bg-primary: white;
+        --bg-secondary: #f6f8fa;
+        --bg-code: #afb8c133;
```
tuurep, line 53: Btw I think we should change this hex to a simpler form with no transparency info (https://www.quackit.com/css/color/values/css_hex_color_notation_8_digits.cfm)

So:

* `#afb8c133` --> `#eff1f3` (color unchanged)

Agreed? :)

static/colors.css
jannis-baum: Yes, good point, I was actually going to suggest the same but adjusting the light mode colors the other way around instead😂 Made more sense to me that thicker borders can be muted and you still see them well whereas thing borders need more contrast to be visible. But of course the other way around it also makes sense, i.e. thinner borders are supposed to be more subtle so they have a more muted color.

If you want, check out how light mode looks if you make the same change the other way around, pick the option you prefer, commit it, approve the PR and then we merge☺️

tuurep: Ah, lets see.

So just to be clear, you think the one I suggested above looks good to you on dark?

jannis-baum: By the way, sorry this is super off-topic here, but could you maybe also check #158? haha. The newest release has an issue where vivify-server has an undefined port when there is no custom config so it crashes on startup. For me #158 fixes it but I have yet to get a confirmation from the people on the corresponding issue🙈 So in case you are quicker and don't mind maybe you could check that ^^

jannis-baum: > So just to be clear, you think the one I suggested above looks good to you on dark?

Yes, I'm fine with either option :) Feel free to decide

static/colors.css
jannis-baum: Yes I guess this works because there is never anything but white behind it? Fine with me, you can also commit this :)

tuurep: Yeah this is something relevant to GitHub but for us feels like being extra obtuse for no reason :D

Ok I'll do a commit once I decide about the borders. Tbh I think I'm liking the regular=muted for dark as well now that I'm looking at it...

But I'll still double check.

tuurep (approved):
Ok, I would go with the muted for header border (etc) option,

And I think it would be simpler if we rename to

* `border-muted`
* `border-regular`

Also partly so that one can reuse these colors in custom styles and it's more descriptive this way

So if you're okay with all this, we can merge :smile:

Sorry for all the pain and suffering we have been through here.

jannis-baum:
Absolutely no reason to apologize! I appreciate your dedication to getting these things right a lot, this is very important to the project. Thank you!

tuurep:
> Absolutely no reason to apologize! I appreciate your dedication to getting these things right a lot, this is very important to the project. Thank you!

Ok! Not always sure if I'm being counter-useful, so good to know :smile:

jannis-baum:
> Ok! Not always sure if I'm being counter-useful, so good to know 😄

Haha no, definitely useful, this is a big contribution to the quality!

I fixed the naming, you forgot to rename `bold` to `muted` in the Markdown styles :)

Now we're good right?

tuurep:
Oh yeah, I forgot to rename the variables.. :facepalm:

Lets be sure that they're correct, careful because I changed the order of them (the old `regular` is not the new `regular`)

jannis-baum:
> Lets be sure that they're correct, careful because I changed the order of them (the old regular is not the new regular)

Ohhh okay haha, then they probably aren't correct. Can you look into it?

tuurep:
Yes, these just had to be flipped

Now it's good!

tuurep:
> I'll be out for today, talk tomorrow! :)

Okay, I see no reason not to merge now so I'll just go ahead and do it :D

Thanks

Issue #146, opened 2024-07-29
Cursor sync offset/more centered (like Vim's `scrolloff`)


tuurep:
From [offtopic discussion](https://github.com/jannis-baum/vivify.vim/pull/17?notification_referrer_id=NT_kwDOAOVvTrQxMTcwODEwODQ1NToxNTAzNjIzOA#issuecomment-2256345587)

Me:

>The way the cursor line is at the very bottom of the page feels pretty disorienting to me when opening
>
>But what I'm getting to with this is that I wish the cursor sync could have a similar option to Vim's [scrolloff](https://vimhelp.org/options.txt.html#%27scrolloff%27), if you know what I mean
>
>But of course it's more complicated because lines aren't a fixed size like in terminal

Jannis:

> Yes definitely. I have also had this thought but haven't followed it to think about how to implement that, haha. We use JavaScript's [`scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) for the scrolling, we could look into that and see how much effort it would be to implement or if that's already a feature there anyways
>
> EDIT maybe also [this](https://stackoverflow.com/a/49860927). Can't look into it in detail right now.

jannis-baum:
Fancy! I have never used this transfer feature before but seems like its perfect for us!

jannis-baum:
@tuurep I think for this we'll need a way to pass config options to the client. Maybe you could adjust

https://github.com/jannis-baum/Vivify/blob/419eb5b3cf6aee67aa2c21dba9f86768cdfc01c8/src/routes/viewer.ts#L85-L86

to just add one property to `window`, e.g.

```javascript
`window.VIV_INFO = {
    "port": "${config.port}",
    "path": "${urlToPath(req.path)}",
    "config": {
        // ...
    }
}`
```

so that we don't pollute it and end up with a million things on `window`.

PR #160, opened 2024-08-06
Improve error reportability

### LINKED ISSUE #159: Improve error reportability
- [x] make sure Vivify can output its version even when everything else fails and that it always completes the startup
  - handleArgs can be called before calling health & starting the server by instead of opening just returning a function that opens and will be called after server is running
  - completion of startup can be handled in `app.ts`
- [x] update bug issue template, include:
  - version
  - system info
  - `vivify-server` output by calling absolute path
  - custom config

jannis-baum:
Close #159

### DIFF
--- a/.github/.markdownlint-cli2.yaml
+++ b/.github/.markdownlint-cli2.yaml
@@ -4,3 +4,6 @@ globs:
 ignores:
   - 'node_modules/**/*'
   - 'tests/**/*'
+  # GitHub issues don't quite abide by Markdown standards (e.g. single line
+  # breaks are reproduced when rendered)
+  - '.github/ISSUE_TEMPLATE/**/*'
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -32,6 +32,8 @@ assignees: ''
 
 > My operating system is:
 
+
+
 > Running `viv --version` outputs:
 
 ```txt
@@ -40,6 +42,12 @@ output goes here
 
 > I installed Vivify in the following way (e.g. name of the package manager, self-compiled or development mode):
 
-> This is what happens when I run `vivify-server` at the absolute installation path instead of `viv` and try to reproduce the problem:
+
+
+> This is what happens when I kill any running instance of Vivify with `killall vivify-server` and then run `vivify-server` at the absolute installation path (e.g. at `command -v vivify-server`) instead of `viv` and try to reproduce the problem:
+
+
 
 > In the following I specify if I use any custom configuration, and if so provide it in its entirety including any additional files referenced there (e.g. CSS or JavaScript):
+
+
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -7,25 +7,39 @@ assignees: ''
 
 ---
 
-### Describe the bug
+[//]: # (Hello and thank you for taking the time to submit a bug report!       )
 
-A clear and concise description of what the bug is.
+[//]: # (To make sure we can resolve this as quickly as possible, please       )
+[//]: # (carefully fill out the information requested on this template.        )
+
+[//]: # (Hint: Lines that look like these are Markdown comments and won't be   )
+[//]: # (      visible on the issue after you submit it so you don't need to   )
+[//]: # (      delete them.                                                    )
+
+### Description
+
+[//]: # (A clear and concise description of what the bug is and what you       )
+[//]: # (expected to happen instead. Feel free to include additional material  )
+[//]: # (such as screenshots if they contribute to the explanation.            )
 
 ### To Reproduce
 
-Steps to reproduce the behavior:
+[//]: # (A list of steps we can follow to consistently reproduce the bug       )
 
-1. Use version ...
-2.
+### Context
 
-### Expected behavior
+[//]: # (Please fill out the following list                                    )
 
-A clear and concise description of what you expected to happen.
+> My operating system is:
 
-### Screenshots
+> Running `viv --version` outputs:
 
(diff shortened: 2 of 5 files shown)


jannis-baum:
Hello @tuurep @Tweekism could one of you take a look at this and leave a review? Thanks! I got motivated to do this after we had quite a lot of back and forth on a bug the other day, I think this should make these things easier in the future.

tuurep:
Whenever I write issues I quite hate these types of templates but that's a personal problem lol

I do see the need but am also thinking we could have a Troubleshooting Readme section

I read all of the template text, and it's good, with 2 worries:

* Is it too verbose
* One of my biggest traps, forgetting to kill the last process (i.e. `killall vivify-server`) might be worth a mention *somewhere*, but could go under some Troubleshooting doc as well?

tuurep:
Do you have an example case where `--version` would hang before, and this fixes? :)

jannis-baum:
> Whenever I write issues I quite hate these types of templates but that's a personal problem lol

Haha yes so do I but I think I got to the point now where I really understand why people have them. Everything in that template are things that I/we would have to individually ask little by little otherwise and this makes the process of figuring out how to reproduce the bug very lengthy and annoying. I'd prefer to just fix it quickly without having to go through a huge conversation ^^

> I do see the need but am also thinking we could have a Troubleshooting Readme section

Yes, this is something we could add in addition. What would you include there?

I still think we need the information in the bug template anyways though because people won't read everything before opening a bug

> Is it too verbose

I tried my best to keep it as short as possible. If you have an idea for how to shorten it without losing information/understandability I'd be happy to improve it based on that!

> One of my biggest traps, forgetting to kill the last process (i.e. killall vivify-server) might be worth a mention somewhere, but could go under some Troubleshooting doc as well?

Good point! I think we could add it to the start of the section where we ask to run `vivify-server` from the absolute path. What do you think?

> Do you have an example case where --version would hang before, and this fixes? :)

You can set `VIV_PORT` to something invalid like `VIV_PORT=hehe viv .`. That made the server crash before which led to not printing `STARTUP COMPLETE`, which led to `viv` hanging indefinitely. Maybe there are other ways of making it crash as well. I hope I am catching everything that can go wrong now

tuurep:
Yeah also it starts making more sense the more popular a repo is

I didn't realize that CONTRIBUTING.md already has a troubleshooting section that includes perhaps the most important one, Node SEA build issue,

and then the others are the ones you have in the template:

* AUR/Brew package vs. compiled: which is running?
* Running with a clean/minimal config
   * (Implied by the last section in the template)

Sooo yeah

> I think we could add it to the start of the section where we ask to run `vivify-server` from the absolute path. What do you think?

This is the best idea here that I think we should do!

My thought was to add a README section until I saw the CONTRIBUTING also has it. Maybe worth extending that with all of the above?

> If you have an idea for how to shorten it without losing information/understandability

I think it's a little wordy, might I say a little too polite if you know what I mean :smile: 

But it's quite fine, we don't need to add anything other than the tip about `vivify-server` unless you wanna.

> You can set `VIV_PORT` to something invalid like `VIV_PORT=hehe viv .`

Ok, nice change! Tested a little and yes, on `0.3.1` it hangs and on this it gives an error where I can figure out what went wrong, even.

You might be aware, but in the case of the Linux build where it segfaults, it still hangs with e.g. `viv .` (only able to tell it segfaults running `vivify-server` directly). Wonder if it's possible to have `viv` exit in that case?

jannis-baum:
So I added the tip about `killall vivify-server` now and I added some more code to `viv` to catch (hopefully) any type of crash, including the segfault. Could you test this?

> My thought was to add a README section until I saw the CONTRIBUTING also has it. Maybe worth extending that with all of the above?

This is (as far as I can tell) the only thing still missing from the stuff you commented. I'll do this next but I'm out of time for today. Let me know if I missed anything else :)

tuurep:
Ok, nice it works

When building purposely with system node and running `viv .`:

```
$ viv .
Fatal: vivify-server crashed. Please use the link below to submit a bug report.

https://github.com/jannis-baum/Vivify/issues/new?labels=type%3Abug&template=bug-report.md
/home/tuure/.local/bin/viv: line 63: 62404 Segmentation fault      (core dumped) nohup vivify-server $@ > "$output" 2> /dev/null
```

Bug report link interesting idea. Hopefully we don't get 100 bug reports for a known issue :smile: But the idea is that the template may point to the correct info.

I would add one newline after the link though.

tuurep:
Hey since we're pretty much finished for the time being, I'll say that next I'm most interested in #146 but I don't know if it's a complicated one.

jannis-baum:
> Ok, nice it works
> 
> When building purposely with system node and running `viv .`:
> 
> ```
> $ viv .
> Fatal: vivify-server crashed. Please use the link below to submit a bug report.
> 
> https://github.com/jannis-baum/Vivify/issues/new?labels=type%3Abug&template=bug-report.md
> /home/tuure/.local/bin/viv: line 63: 62404 Segmentation fault      (core dumped) nohup vivify-server $@ > "$output" 2> /dev/null
> ```

Oh, very interesting that you get the segmentation fault info now as well! That wasn't the case before, was it?
 
> Bug report link interesting idea. Hopefully we don't get 100 bug reports for a known issue 😄 But the idea is that the template may point to the correct info.

Haha, yes. I hope people will read the stuff on the bug report. I added some more info to that message, what do you think?

> I would add one newline after the link though.

I added the newline, good call.

jannis-baum:
> Hey since we're pretty much finished for the time being, I'll say that next I'm most interested in #146 but I don't know if it's a complicated one.

Sounds great!

tuurep:
> Oh, very interesting that you get the segmentation fault info now as well! That wasn't the case before, was it?

Yeah it would just hang and I couldn't get that info from it

tuurep (approved):
Okay I think we can merge :smile:

Issue #131, opened 2024-07-26
Task list alignment issue


tuurep:
From https://github.com/jannis-baum/Vivify/pull/70#issuecomment-2226902971

> Btw quite disturbing:
>
> ![image](https://github.com/user-attachments/assets/6682fc10-6168-4bcc-b13a-016013e7676f)
>
> How the checkbox isn't aligned to the text. I might look into that next 😄

PR #162, opened 2024-08-12
SVG Icons (octicons)

### LINKED ISSUE #102: SVG icons
- for directory list to replace current 📁 and 📄
- for up/back button after #82 is merged

tuurep:
Close #102

### DIFF
--- a/src/parser/parser.ts
+++ b/src/parser/parser.ts
@@ -6,6 +6,11 @@ import config from '../config.js';
 import renderNotebook from './ipynb.js';
 import renderMarkdown from './markdown.js';
 import { globSync } from 'glob';
+import octicons from '@primer/octicons';
+
+const dirIcon = octicons['file-directory-fill'].toSVG({ class: 'icon-directory' });
+const fileIcon = octicons['file'].toSVG({ class: 'icon-file' });
+const backIcon = octicons['chevron-left'].toSVG({ class: 'icon-back' });
 
 export type Renderer = (content: string) => string;
 
@@ -14,9 +19,11 @@ const pathHeading: Renderer = (path: string) => `# \`${path.replace(homedir(), '
 function wrap(contentType: string, content: string, linkPath?: string): string {
     let link = '';
     if (linkPath) {
-        link = `\n<div id="top-nav">\n<a id="top-nav-up" href="${pathToURL(linkPath)}"> ${
-            pbasename(linkPath) || '/'
-        }</a>\n</div>`;
+        link = `<div id="top-nav">
+                    <a id="top-nav-up" href="${pathToURL(linkPath)}">
+                        ${backIcon}${pbasename(linkPath) || '/'}
+                    </a>
+                </div>`;
     }
     return `<div class="content-${contentType}">${link}\n${content}</div>`;
 }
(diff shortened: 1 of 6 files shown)


tuurep:
I decided to do this in 2 steps:

- [x] Just add the icons + their colors
- [ ] svg-to-text alignment
    * Hopefully we can use the same solution for #131

tuurep:
Notes:

* Made sure there's no whitespace between svg and link text, because it looks awkward when the space gets the link underline
* I removed some `\n`s from the `#top-nav` and am not fully sure if they were necessary there but I'm trying to hopefully use CSS for these sort of things

src/parser/parser.ts
```diff
@@ -14,9 +19,11 @@ const pathHeading: Renderer = (path: string) => `# \`${path.replace(homedir(), '
 function wrap(contentType: string, content: string, linkPath?: string): string {
     let link = '';
     if (linkPath) {
-        link = `\n<div id="top-nav">\n<a id="top-nav-up" href="${pathToURL(linkPath)}"> ${
-            pbasename(linkPath) || '/'
-        }</a>\n</div>`;
```
tuurep, line 19: Some literal `\n`s removed here, @Tweekism pls warn me if they were necessary

My thought process is that when I get the icon alignment right, aligning the divs and such are gonna use CSS (margins/paddings) anyways

tuurep:
@jannis-baum I'm proposing new colors for dark theme dir and file icons:

image
[](url)

How's that look? (if you can ignore the alignment :smile:)

dir: `#59739c`
file: `#989898`

For back button I'm suggesting to reuse `--text-secondary`

image

image

(todo: alignment)

src/parser/parser.ts
Tweekism: the \n are for formatting of the actual html source, not the rendered output.

tuurep: Oh I see now. Since I added some actual linebreaks in the string it should achieve a similar neatness of the HTML I suppose. But other parts of the code don't have this goal in mind and maybe it's not important :smile:

tuurep:
Ready for review

image

backbutton-after

jannis-baum:
Hey @tuurep thank you, this looks cool! I will look at it as soon as I can, might still be a few days though. Talk soon!

jannis-baum (commented):
I took a look now, looks great overall, such a big improvement over what we currently have! I have two comments, one on the colors and another on customizability of the icons:

### Colors

I felt like the colors were a bit much on dark mode so I tried changing the directory color to the same as the file color. Then I noticed it looked weird because the links were still blue so I figured I'd try out how it looks with the links in a neutral color for both color schemes. I committed my suggestion, let me know what you think! This is what it looks like while hovering on the `README.md`:

image image

### Customizable icons

Do you think it would make sense to allow people to pick alternative icons? We could add a config option for that. If we do, the important thing would be to check that the icon actually exists and catch it if not. But I also definitely see the point of not making them customizable until someone asks for it as this adds complexity to maintaining Vivify.

tuurep:
Yeah, love the link color change. I also prefer that but didn't think to suggest it.

Gray icon colors look good on dark theme, approved

> Do you think it would make sense to allow people to pick alternative icons?

Yeah absolutely, but also I was thinking it's kind of already possible with custom stylesheets with something like:

```css
.icon-directory > path {
   d: path ('foo foo foo foo foo foo foo foo etc')
}
```

or save an svg file in my vivify conf dir and point to it

(dunno how accurate the above are, I could test this as well)

So I think it might already be well customizable, but does require some knowledge of CSS (but so does all other theming)

What do you think?

Also, I think we can merge but I think this is pending 1 approval from either you or tweekism :)

jannis-baum:
Oh nice, I wasn't aware of that CSS feature! Cool. Yes, totally enough I'd say. We can point people to it if they ask for it. I'll approve and you can merge it :)

tuurep:
Yeah can confirm this is possible, for example:

```css
/* https://primer.style/foundations/icons/feed-star-16/ */
.icon-directory > path {
    d: path("M8 16A8 8 0 1 1 8 0a8 8 0 0 1 0 16Zm.252-12.932a.476.476 0 0 0-.682.195l-1.2 2.432-2.684.39a.477.477 0 0 0-.266.816l1.944 1.892-.46 2.674a.479.479 0 0 0 .694.504L8 10.709l2.4 1.261a.478.478 0 0 0 .694-.504l-.458-2.673L12.578 6.9a.479.479 0 0 0-.265-.815l-2.685-.39-1.2-2.432a.473.473 0 0 0-.176-.195Z");
}

```

image

However we might be quite reliant that the SVG used is exactly 16x16 pixels, but maybe that's tweakable on the `svg` tag as well?

jannis-baum:
Great, thanks for checking it! I think as long as it's possible it's fine for now :) If it doesn't work for whoever will ask for it in the future we can reconsider

PR #165, opened 2024-08-26
Mermaid Support

### LINKED ISSUE #132: Mermaid Support
Would be awesome to see mermaid graphs displayed on the preview.

jannis-baum:
Close #132 

Mermaid is incredibly annoying to run server-side and it seems they really don't want people to do this. So I ended up making it run client-side by symlinking the Node module into the `static/` directory. And turns out the same thing works for Katex, so I did it there as well which cleaned up our repo quite nicely

The client-side parsing on the other hand is very easy to implement, I just followed [this example](https://mermaid.js.org/config/usage.html) and created the `class="mermaid"` containers analogously to how we parse Dot/Graphviz.

One consideration might be to let people have [Mermaid configuration](https://mermaid.js.org/config/schema-docs/config.html) but it seems that there is quite a bit in those configuration options that can break things so I would hold off on it until someone asks for it.


tuurep (approved):
Ok looks great, other than the colors on dark theme

image

But I'd say the same goes for "Dot" so if that can be changed as well, I'll give this an 'Approve' in case we want to do those as a separate effort?

jannis-baum:
Just for keeping protocol here as well: I added automatic theming, it will now use `dark` for dark mode and `default` for light mode. Drawback is that we can't dynamically change this, the page has to be reloaded. Making it change dynamically would be quite hacky so I'd avoid it (at least for now)

jannis-baum:
Update: I changed it so that the page refreshes now when the theme changes and there is Mermaid content

tuurep:
Nice approach!

I am quite unable to test this as my system doesn't have a concept of `prefers dark/light` I think..

Or if you know an alternative way to test if this works, let me know.

(I've had a habit to rebuild the whole thing whenever I make *any* changes, including CSS. Should probably get better habits.)

PR #176, opened 2024-08-30
Improve install robustness

### LINKED ISSUE #175: Improve install robustness

jannis-baum:
Close #175

- `viv` now uses `vivify-server` from directory its in to prevent confusion from multiple installs
- i used https://www.shellcheck.net on the script and realized there was a bug that didn't let us open targets with spaces, this is now fixed
- i dropped support of super old non-POSIX compatible shells that don't understand the `$()` syntax so that the script is more readable


tuurep (approved):
Looks nice

I made sure nothing breaks and also if I deliberately mismove or remove permissions from the server executable, the right message outputs

PR #180, opened 2024-09-05
Support .config directory

### LINKED ISSUE #179: Support .config directory

jannis-baum:
Close #179

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -65,8 +65,8 @@ const config = ((): Config => {
         if (!fs.existsSync(cp)) continue;
         try {
             config = JSON.parse(fs.readFileSync(cp, 'utf8'));
-            break;
         } catch {}
+        break;
     }
 
     // revert to default config if no config found


tuurep:
Nice, thanks

I have some doubts that the precedence is working as intended, assuming that the `~/.config/vivify/config.json` should be applied over `~/.vivify/config.json`, when I test with both it looks like the `~/.vivify/config.json` is being applied. Could you verify?

tuurep:
Darn it. The classic "I have an AUR package installed and didn't realize" -case. Let me try again :smile:

tuurep:
Alright I gave it a good with each supported path and works great + I like the flexibility of it

The only thing is that if I have an empty `config.json` for example in the `~/.config/vivify/` path, I'd expect it to act as a 'default config' but when I have a `~/.vivify/config.json` at the same time, that gets applied instead

What do you think about that?

jannis-baum:
Currently, the first path that has *a valid JSON object* will be used. An empty file is not valid JSON. We can change it to try using the first path with a file no matter what. Do you think that's better?

tuurep:
Hmm. Well, that would match my expectation of what should happen, but is that a small and easy change?

jannis-baum:
> Hmm. Well, that would match my expectation of what should happen, but is that a small and easy change?

Yes haha: https://github.com/jannis-baum/Vivify/pull/180/commits/bc70a8cdc9ef3462410e4500cb7503fc06d7b80a. Can you verify that it works how you would expect now?

tuurep (approved):
Lol it's the smallest change of all time :tada: 

Yes I think it's quite perfect now, thanks!

PR #196, opened 2024-12-16
Bug: `:n` file suffix doesn't scroll if browser (firefox) wasn't already open

### LINKED ISSUE #193: Bug: `:n` file suffix doesn't scroll if browser (firefox) wasn't already open

### Description
When using the `:n` CLI file suffix to scroll to a certain line, if firefox wasn't already open, the viewer doesn't scroll. 

(`viv` launches firefox).

Practical use case where this matters: using the [vim-plugin](https://github.com/jannis-baum/vivify.vim), you want to render the markdown on the line you're currently looking at in Vim.

[//]: # (A clear and concise description of what the bug is and what you       )
[//]: # (expected to happen instead. Feel free to include additional material  )
[//]: # (such as screenshots if they contribute to the explanation.            )

### To Reproduce

1. Have firefox set as default browser.
2. With no browser running:
  - `$ viv README.md:90` (example: this repo's README.md)
3. Doesn't scroll after opening the browser and viewer (opens at the top)

[//]: # (A list of steps we can follow to consistently reproduce the bug       )

### Context

[//]: # (Please fill out the following list                                    )

Arch Linux
```txt
vivify-server v0.6.3.r0.g5d07d16-1-aur
```
[`vivify-git`](https://aur.archlinux.org/packages/vivify-git) (AUR)

jannis-baum:
Close #193

### DIFF
--- a/static/client.js
+++ b/static/client.js
@@ -15,10 +15,6 @@ function viv_scrollTo(value) {
 
 const ws = new WebSocket(`ws://localhost:${window.VIV_PORT}`);
 
-ws.addEventListener('open', () => {
-    ws.send(`PATH: ${window.VIV_PATH}`);
-});
-
 ws.addEventListener('message', (event) => {
     const fields = event.data.toString().split(': ');
     if (fields.length < 2) return;
@@ -39,4 +35,8 @@ ws.addEventListener('message', (event) => {
             console.log(value);
             break;
     }
+});
+
+ws.addEventListener('open', () => {
+    ws.send(`PATH: ${window.VIV_PATH}`);
 });


jannis-baum:
Hey @tuurep could you confirm that this fixes it? Turns out the bug made sense haha, it seems Firefox runs the web socket `open` event listener immediately while Chromium and Safari seem to finish the JS file first which is why the order didn't matter there. Or they're just slow. Anyways, for me it works like this on Firefox as well.

tuurep:
Initial test:

On firefox (clean profile), this doesn't fix it for me

Also I noticed that firefox-developer-edition doesn't have this problem (even before this PR), but not sure how relevant this is :)

If the problem is in how fast the browser starts up, one thing I notice is that ff-dev-edition loads a little bit faster.

jannis-baum:
I tried this out a bit again and I actually also still get the problem around 1/20 times. The other times it works. Can't figure out what causes it, it seems completely random.

Either way, this looks a lot like Firefox's fault to me now to be honest, especially since it's so inconsistent and works with other browsers. Not sure what else we could do other than introduce some random delay of sending the scroll message but that would worsen UX for everyone else so I'd rather not do it. Do you have another idea? Or do you know when we could expect Firefox to be upgraded to whatever the dev version is on now that fixes it?

tuurep:
Hey, inspired by your 1/20 test on firefox, I tried the same on firefox developer edition and it can similarly happen randomly for me on firefox developer edition as well.

Yeah, I have no clue what to do about it but if something comes up I'll let you know.

jannis-baum:
Since there doesn't seem to be anything we can do, let's merge & close this?

tuurep (approved):
Yeah sounds good

PR #197, opened 2024-12-17
Mermaid charts convert to text on live reload until page refresh

### LINKED ISSUE #194: Mermaid charts convert to text on live reload until page refresh

### Description

On live reload (i.e. modifying the mermaid code block) it gets rendered as a regular markdown code block instead of a Mermaid chart.

[//]: # (A clear and concise description of what the bug is and what you       )
[//]: # (expected to happen instead. Feel free to include additional material  )
[//]: # (such as screenshots if they contribute to the explanation.            )

### To Reproduce

Have a document such as

````
```mermaid
flowchart LR
    A[foo]-->B[bar]
```
````

which renders as (screenshot):

image

Start editing the mermaid block, such as

````
```mermaid
flowchart LR
    A[foo]-->B[bar]
    B-->C
```
````

It renders as this (screenshot):

image

until manual page refresh:

image

[//]: # (A list of steps we can follow to consistently reproduce the bug       )

### Context

[//]: # (Please fill out the following list                                    )

- Firefox 133.0
- Arch Linux
- [`vivify-git`](https://aur.archlinux.org/packages/vivify-git) (AUR)
  - `vivify-server v0.6.3.r0.g5d07d16-1-aur`

jannis-baum:
Close #194

Main change is calling `mermaid.run(...)` in the `UPDATE` message callback. To make that work I had to move around a bunch of stuff, resulting in all our client-side stuff now also being a more modern JS module instead of a plain script. And hopefully more readable.


tuurep:
Alright

Context: I told Jannis I see an annoying visual flicker which is best seen in a video recording. Now I finally got around to recording this. Other than that, the fix is very nice.

https://github.com/user-attachments/assets/7f50bedd-f077-47ca-879f-6d9a5d4a4abe

So whenever there's a mermaid chart, on every keystroke it has to kinda reposition itself on the page, even if not editing the mermaid part itself. Editing those large h1 elements below shows it best.

Note, if there's a syntax error, this doesn't happen, which is nice:

https://github.com/user-attachments/assets/e85b38f3-1144-44a7-8f6c-86dcb57bbf95

jannis-baum:
Yeah the flicker is because Mermaid *really* wants to be rendered client-side. I couldn't find a way to render it server-side without introducing a ton of really ugly hacks and faking a DOM, which doesn't seem worth it to me just because of Mermaid. You can also check again if you find a good way of rendering server-side, otherwise I would suggest leaving it like this and merging.

tuurep (approved):
Tried to look for any relevant issues on the server- vs. client-side rendering, but couldn't see any.

I'd say we should merge this, and could do a release at this point.

Issue #202, opened 2025-01-24
Code Copy Buttons


Tweekism:
Feature request to add copy code buttons to code blocks.

I'll add a draft PR soon with work done so far.

![Image](https://github.com/user-attachments/assets/2e3ef810-424d-4832-8223-87351de06058)

jannis-baum:
Closed in #203

PR #203, opened 2025-01-24
Add code copy buttons using 'markdown-it-copy' plugin

Tweekism:
Issue #202 

Hi all,

I'm working on adding "copy code" buttons like seen on GitHub, Microsoft docs, stack overflow etc.

Feedback so far:-
- Have the button be invisible until the `<pre>` block is hovered over
- Have the default visibility be configurable

Default visibility can be toggled back on in custom css.

One thing I'm still struggling with is the alignment issue. @jannis-baum I tryed your suggestion, but turns out I had already set the `<pre>` element with the scroll `overflow` to relative and then set the button to `absolute` with a `top`, `right` offset. However, while that sets the starting position correctly, it then scrolls along with the overflow box (see blow for example). 

If I wrap the `<pre>` in a containing `<div>` and set the the button relative to the `<div>` then it works as expected, but the `<pre>` blocks are created by the markdown plugin and I don't know how or if that is the right approach.

alignment

A comparison of the change

div-wrapper

So looking for feedback / ideas on that.

I've pushed the PR branch here in-case anyone wanted to try it out.


Tweekism:
As a side note...

The plugin has a "success" popout thing after you click the button. Out-of-the-box it shows up for 2 seconds then disappears. The javascript for that is contained in the plugin but with some css you can set it to fade out when you leave the button.

At first I thought option 1, where it lingered around, looked kinda weird, but after using it for a few days i noticed your eyes tend to move off to where you are pasting to really see it. In any case, I also tryed option 2 and at first I liked it but in fact having it fade away instantly if you are moving fast might be weirder.

But I don't know, let me know what you think is a more sane default. Option 1 is simpler. Option 2 is a couple extra lines of css (lines 224-226 added in markdown.css). Option 2 is currently set but you can set opacity on line 204 back to `1` to turn off the effect.

Of course either could be achieved in custom css.

option1

option2

jannis-baum:
Hello! How is it going on this? When testing with `yarn dev` this looks great to me but when I try to install it and run the binary I get a weird error that I don't understand:

```plain
node:fs:563
  return binding.open(
                 ^

Error: ENOENT: no such file or directory, open '/Users/jannisbaum/.bin/lib/code.js'
    at Object.openSync (node:fs:563:18)
    at Object.readFileSync (node:fs:447:35)
    at 3928 (build/bundle.js:2:448835)
    at __webpack_require__ (build/bundle.js:2:4588852)
    at 3226 (build/bundle.js:2:3374025)
    at __webpack_require__ (build/bundle.js:2:4588852)
    at 6421 (build/bundle.js:2:3402535)
    at __webpack_require__ (build/bundle.js:2:4588852)
    at 2466 (build/bundle.js:2:1517545)
    at __webpack_require__ (build/bundle.js:2:4588852) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/jannisbaum/.bin/lib/code.js'
}

Node.js v23.7.0
```

Do you know what that `code.js` file is supposed to be and do you also get this error?

Tweekism:
Yeah I discovered this the day after posting, happens to me too. I don't know why it is looking for code.js, must be the npm `markdown-it-copy` package. I'll look into it.

Tweekism:
Hi all,

So finally got to work on this a bit, I have re-implemented the copy code buttons to not use any mdit plugins (they were all a bit crap or over engineered for our needs). I did use a clipboard package, as to handle that properly across all browsers is a little involved, this npm package [clipboard](https://www.npmjs.com/package/clipboard) seems very popular and well behaved.

Let me know what you think.

static/colors.css
```diff
@@ -19,6 +19,13 @@
     --icon-file: var(--text-secondary);
     --icon-back: var(--text-secondary);
 
+    --copy-button-fg: #aaa;
+    --copy-button-bg: #333;
+    --copy-success-fg: #ddffcc;
+    --copy-success-bg: #5c993d;
+    --copy-fail-fg: #ffcccc;
+    --copy-fail-bg: #993d3d;
```
jannis-baum, line 27: I don't know about @tuurep's opinion for light mode but for dark mode I wouldn't really need different colors depending on the state; the icon change is enough for me. The current colors pop too much which is not necessary because the user is most likely already looking there anyways after clicking the button and breaks the minimalistic look of the viewer a bit

jannis-baum (changes requested):
Looks great overall & works! Thanks a lot! Check out the small comments and let's see what Tuure says as well :)

static/colors.css
```diff
@@ -19,6 +19,13 @@
     --icon-file: var(--text-secondary);
     --icon-back: var(--text-secondary);
 
+    --copy-button-fg: #aaa;
+    --copy-button-bg: #333;
```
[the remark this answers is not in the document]
Tweekism: Well actually the fg here is actually analogous to the primary and secondary text, its just a custom svg glyph instead of a letter. So maybe it should be --text-tertiary. Or even just use the existing --text-primary since it is the same?

image

static/colors.css
Tweekism: I can remove it entirely, or i can set the default values to be `...: var(--...as above etc...)` leaving looking the same but keeping it configurable in custom css.

tuurep:
Thanks, it works great

We chatted about:

- Animation feels slightly too slow, change it from 0.2s to ~0.1s by default
- As Jannis said, the icon change is plenty enough for copy-success, and looks great without a green success-bg color

static/colors.css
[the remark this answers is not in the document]
Tweekism: Simplified color scheme to use tertiary grays, users can add colors back in custom css if desired

static/colors.css
Tweekism: Simplified color scheme to use tertiary grays, users can add colors back in custom css if desired

jannis-baum (approved):

PR #205, opened 2025-02-09
Upgrade dependencies

### LINKED ISSUE #204: Upgrade dependencies

jannis-baum:
Close #204


tuurep:
Yes looks like everything's OK, just made sure everything in `tests/rendering/*.md` still works.

I think it can be merged unless something needs special attention.

tuurep (approved):

docs/upgrading-dependencies.md
```diff
@@ -0,0 +1,42 @@
...
+6. Run `yarn-deduplicate` and then `yarn` to keep only the newest required
+   versions of dependencies and hopefully address all security issues the
+   [Dependabot](https://github.com/jannis-baum/Vivify/security/dependabot) had
+   found.
+7. Repeat *step 2* a final time, open a PR and ask another maintainer for a
+   review.
+
+## Reviewing upgrade PRs
+
+You can following along *step 2* above to check things are working and leave
```
tuurep, line 41: *follow

jannis-baum: Fixed in https://github.com/jannis-baum/Vivify/pull/205/commits/031727d7dd64af5109a0d339db0c5e8b18664bce

tuurep:
`upgrading-dependencies.md`

This is great, I could reproduce it on Arch and I can see how doing this without this memo is a pain each time.

Some problems with `yarn-deduplicate`:

1. Isn't a global npm/yarn install a bad practice?
  - though I'm mostly thinking about the python `pip` equivalent
  - Could we have like local dev dependencies with `yarn-deduplicate` and anything in the future we might need?
2. If installed with `yarn global add`, the bin goes to `~/.yarn/bin` on my machine which isn't on the `$PATH` by default
  - apparently yarn bin can be in a number of different places, but you can get your path with `yarn global bin`
  - if we could install `yarn-deduplicate` locally, this would be a non-issue

jannis-baum:
Nice, very good idea! I made the adjustment. Can you confirm it works?

tuurep:
Works, thanks!

PR #213, opened 2025-04-13
Set Notebook ANSI colors

### LINKED ISSUE #101: Set Notebook ANSI colors

jannis-baum:
Close #101

Sets up 16 ANSI colors as in my jellyfish theme for dark mode (I know they're not optimal either haha but at least now easily configurable), and GitHub's notebook viewer for light mode:

image
image

One caveat is that quite a few of the colors (e.g. function names, string literals) used in the exception traces are actually from the 256 ANSI colors so while it's better than before, it's still hard to read in dark mode :(

image

I'll note that the above output was produced by using Jupyviv. If I run the same thing in a Jupyter Lab instead and then render the notebook with Vivify, it looks like this:

image

For reference, this is how the Jupyter Lab-executed notebook renders on Vivify before this PR:

image

Seems like Jupyter Lab somehow makes the kernel use just the 16 colors. I'll check if I can somehow do this with Jupyviv as well, maybe by setting the `TERM` environment variable or something (→ https://github.com/jannis-baum/Jupyviv/issues/6)

Light mode looks good & readable, both when using the 16 colors as well as when using the 256 colors from Jupyviv

EDIT: Just realized that the Jupyter Lab version also still uses colors other than the 16. You can see it in the dark green `def`s for example, that color isn't one of the ones above

### DIFF
--- a/src/parser/ipynb.ts
+++ b/src/parser/ipynb.ts
@@ -16,6 +16,7 @@ import { Renderer } from './parser.js';
 import { AnsiUp } from 'ansi_up';
 
 const ansiup = new AnsiUp();
+ansiup.use_classes = true;
 const renderAnsi: Renderer = (content: string) => ansiup.ansi_to_html(content);
 
 function joinMultilineString(str: MultilineString): string {
--- a/static/colors.css
+++ b/static/colors.css
@@ -1,5 +1,6 @@
 /* --------------------------------------------------------------------------
  * DARK MODE ---------------------------------------------------------------- */
+/* from https://github.com/jannis-baum/dotfiles/blob/main/lib/color-schemes/jellyfish.yaml */
 
 :root {
     --bg-primary: black;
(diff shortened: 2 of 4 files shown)


tuurep:
Testing this, why is it that my colors look way different when viewing the `tests/rendering/notebook.ipynb`?

image

```
$ viv -v
vivify-server v0.7.0-3-ge0aa9f7
```

jannis-baum:
Hm that's weird. Maybe you still had the old `vivify-server` running? Just checked this again and I do get the same colors as the screenshot

tuurep:
Ok, I never quite found out why I saw wrong colors yesterday, but now I see these:

image

So, looks good, but the thing that catches my attention the most is the readability of the cyan here:

image

Could be worse. Did we want to tweak anything further or what's the plan?

jannis-baum:
The light mode colors are just taken out of the GitHub light theme so I think I'd leave those as they are.

If you have any suggestions for dark mode colors and/or the way I implemented this we can make adjustments there, if not I think we can just merge :)

tuurep:
Yeah I doubt I have anything useful to add,

as long as this is the current way the dark mode looks (from your opening post):

image

That looks really good (apart from the horrid light gray on yellow error, but I've had a lot to say about that :smile: [root cause](https://github.com/ipython/ipython/issues/13446))

For what it's worth, I would be configuring the contrast of some of those (light theme) colors if I were to use Viv + jupyter-notebooks, which I haven't been doing yet but will definitely check out the Jupyviv at some point.

So about implementation, looks like it's easily themable, right? As long as that's the case I can approve.

tuurep (approved):

tuurep:
Well I gotta see it to believe it, this is my screenshot:

image

Yeah I missed the part about JupyterLab. The blue and red look awful on the dark mode as above,

Was there anything we could do about it?

jannis-baum:
Yes, as you saw the horrible colors are because the kernel outputs 256 colors and those are not theme-able with the ANSI plugin we use. Would also be way out of scope to change all 256 colors though. I think from Vivify's side setting the 16 is as good as it's going to get and then I'll see if there is some way to make the kernel output base16 colors in Jupyviv.

> So about implementation, looks like it's easily themable, right? As long as that's the case I can approve.

Yes definitely, comes down to just setting the color variables like all other colors :)

---

Thanks! Going to merge this now

PR #216, opened 2025-04-30
Render clickable wiki-links, i.e. [[another-note]]

### LINKED ISSUE #215: Render clickable wiki-links, i.e. [[another-note]]
It's quite common for note taking apps to insert links using the wiki-link format [[like-this]], which would link to a note in the same dir called `like-this.md`.

It'd be nice if they were properly rendered as a clickable links.

jannis-baum:
Close #215

### DIFF
--- a/src/parser/wiki-links.ts
+++ b/src/parser/wiki-links.ts
@@ -1,4 +1,5 @@
 import type MarkdownIt from 'markdown-it';
+import { basename as pbasename } from 'path';
 
 export default function wikiLinkPlugin(md: MarkdownIt): void {
     md.inline.ruler.before('link', 'wiki_link', (state, silent) => {
@@ -17,7 +18,8 @@ export default function wikiLinkPlugin(md: MarkdownIt): void {
 
         if (!silent) {
             const content = state.src.slice(start + 2, end - 2);
-            const href = content.indexOf('.') > -1 ? content : content + '.md';
+            const hasExtension = pbasename(content).indexOf('.') > -1;
+            const href = hasExtension ? content : content + '.md';
 
             // Create link tokens
             const token = state.push('link_open', 'a', 1);


tuurep:
Yes this works for the case requested in the issue as far as I can see.

It's totally hand-written, right? I'm not really familiar enough with how to write these to verify the code enough, but on testing, works with:

- `md` file on the same dir: `[[foo]]` => `/bar/baz/foo.md`
- relative path:
  - `[[../qux]]` => `/bar/qux.md`
  - `[[garply/grault]]` => `/bar/baz/garply/grault.md`

Doesn't work with absolute paths, but maybe that's not the point either.

Obsidian has some docs like this (calling it "internal links"):

https://help.obsidian.md/links

Looks like there can be some more advanced syntax there, but what we have here looks good to add and could be extended on request?

tuurep:
Oh btw, noticed the `[[toc]]` (toc-plugin) uses the exact same syntax,

Verified that then you cannot have a file called `toc.md` and try to link to it this way :smile: 

This is probably fine and the best we could hope about that.

tuurep:
Still curious about one thing:

When writing `[[foo.md]]` (including the extension explicitly), would this work in other editors?

On our implementation it links to `/bar/baz/foo.md.md`

I definitely don't like that :smile:

jannis-baum:
Thanks a lot as always, great job looking at more different cases and researching!

I made the adjustment for the explicit file extensions. It'll result in maybe slightly weird behavior for someone who has a `some.file.md` and wants to link to it as `[[some.file]]` which won't work. But I guess if you do that it's fair that you have to write `[[some.file.md]]` haha.

> It's totally hand-written, right?

Yes, I couldn't find any markdown-it plugin that does that (checked https://www.npmjs.com/search?q=markdown-it-wiki-links) so went for the manual implementation while looking at [the native inline parsers](https://github.com/markdown-it/markdown-it/tree/master/lib/rules_inline) for reference and having DeepSeek help haha.

> Obsidian has some docs like this (calling it "internal links"):
>
> https://help.obsidian.md/links
>
> Looks like there can be some more advanced syntax there, but what we have here looks good to add and could be extended on request?

Yes good idea, let's do it like that.

---

Could you check again and approve if it's good? :)

tuurep:
Nice, works well!

One last thing I wanted to try was filenames with spaces: works

> It'll result in maybe slightly weird behavior for someone who has a `some.file.md` and wants to link to it as `[[some.file]]` which won't work. But I guess if you do that it's fair that you have to write `[[some.file.md]]` haha.

This also applies to relative (parent) paths, which now require the `.md`.

Do we care about that? :smile: I'm not sure.

jannis-baum:
Good idea! And yes, easy fix hehe, should be good now.

tuurep (approved):
Alright, all works great

PR #219, opened 2025-06-05
Execute updated `<script>` tags

### LINKED ISSUE #218: Execute updated `<script>` tags
Another thing needed for Jupyviv but might also make some other stuff in the future easier:

Whenever Vivify soft-reloads (i.e. content is changed without the browser actually reloading the page), content that relies on `<script>`s being executed to display properly doesn't display properly because those `<script>`s are not executed by being set as `.innerHTML`, meaning we have to hard-reload to see it.

The PR for this issue changes that so that `<script>`s are executed on soft-reloads. One way to try it out is with a Notebook[^1] with the following code executed:

```py
import plotly.io as pio
pio.renderers.default = 'notebook'

import plotly.express as px
fig = px.scatter(x=[0, 1, 2, 3, 4], y=[0, 1, 4, 9, 16])
fig.show()
```

Opening that Notebook with Vivify and soft-reloading it, e.g. with the following

```sh
curl \
    -X POST \
    --data '{"reload": 1}' \
    '-H' 'Content-type: application/json' \
    localhost:31622/viewer/path/to/your/file.ipynb
```

will make the plot vanish before this PR and persist after this PR.

[^1]: I can't attach the `.ipynb` file here and also don't want to upstream it to the repo because the output HTML is huge and would make Vivify's main language into Jupyter Notebook🙃🥲

jannis-baum:
Close #218

### DIFF
--- a/static/client.mjs
+++ b/static/client.mjs
@@ -33,8 +33,27 @@ ws.addEventListener('message', async (event) => {
 
     switch (key) {
         case 'UPDATE':
-            document.getElementById('body-content').innerHTML = value;
+            const container = document.getElementById('body-content');
+            container.innerHTML = value;
+
             await mermaid.run({ querySelector: '.mermaid' });
+
+            // find & execute <script> tags in updated content to prevent
+            // having to hard-reload page for added JS
+            const scripts = container.querySelectorAll('script');
+            scripts.forEach((script) => {
+                // create new script element that is copy of original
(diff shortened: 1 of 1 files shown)


tuurep (approved):
Just made sure I could reproduce the case on the issue: https://github.com/jannis-baum/Vivify/issues/218#issue-3120932345

It works

PR #223, opened 2025-06-08
Ask for package contributions

### LINKED ISSUE #222: Ask for package contributions

jannis-baum:
Close #222


tuurep (approved):
Yeah looks good!

Unnecessary opinion: I think asking for a star is a little distasteful (I feel the same way about youtubers asking for like and sub)

But it's normal I guess, we can put that in if you don't share my view :smile:

jannis-baum:
Hm yes I agree haha but (1) I see it on so many repos, (2) we have more downloads on releases than stars even though the AUR package doesn't even use the download, and (3) it's relatively common to see people open issues here who seem to use and like Vivify but didn't star it.

So that's why I decided against my shared dislike for asking for it haha. What do you think? If you're not convinced we can drop it

tuurep:
It doesn't convince me simply because: why do we need stars? 😄 I think Vivify is doing fantastic whether it has 0 stars or 70 stars.

If you let me decide, we should drop it.

Issue #141, opened 2024-07-29
Obsidian style callouts


Tweekism:
As per https://github.com/iamcco/markdown-preview.nvim/discussions/682

Feature requested on iamcco plugin by 2 recent users.

OP @Praczet
> Hi,
> 
> Firstly thank you for this plugin I really like it.
> 
> I would like to ask if it will be possible to add support for the Obsidian's callouts (markdown-it-obsidian-callouts)
> 
> Best regards

jannis-baum:
Is this a full superset of GitHub alert blocks? I.e. if we include this, will the GitHub blocks still work?

Tweekism:
That's what I was looking at. How deep do you want to go implementing stuff like this, obsidian on its own is a whole massive rabbit hole.

jannis-baum:
> obsidian on its own is a whole massive rabbit hole.

Yep, exactly. That's why I would only do it if people ask for it explicitly. Also, I'm not familiar with Obsidian, but when I opened their page for a second it looked like a huge thing including the editor and viewer and all that so I don't know if people even have files from that laying around that they would want to look at in Vivify/edit in Vim

Tweekism:
Then lets close for now as not planned? and revisit if there's demand for it.

jannis-baum:
Yes sounds good! We can reopen if @ Praczet stops by and/or someone else wants it as well.

pidgeon777:
I'm also very interested, hopefully it will be implemented one day 🙂.

gaardhus:
Also looking for this feature 👍

tuurep:
Fair to say it's a popular request, reopening

But I have the same question as @jannis-baum above:

Is this similar to [Github Style Alerts](https://github.com/antfu/markdown-it-github-alerts)?

The biggest functional difference I can see is [foldability](https://help.obsidian.md/Editing+and+formatting/Callouts#Foldable+callouts)

And another one is [nestability](https://help.obsidian.md/Editing+and+formatting/Callouts#Nested+callouts), it apparently doesn't work with our github alert plugin, but I'd be interested in 'fixing' that.

i.e.:

![image](https://github.com/user-attachments/assets/a24e8857-17c0-4c90-bbaa-d840a1dd9aad)

Other than that it looks like something that could be achieved through [theming](https://github.com/jannis-baum/Vivify/blob/main/docs/customization.md#configuration-file) the github alerts.

tuurep:
[...]

PR #233, opened 2025-08-20
Issue/141 obsidian style callouts

tuurep:
Issue: #141

I'd accidentally removed this branchfrom my fork, which adds everything other than custom icons support, which is the main thing this is stuck on. Let's re-add, this time opening a draft PR.

Problems:

Icons need to be able to be able to be specified 3 different ways:
1. octicons
2. inline svg
3. svg from file

The idea is to do that configuration in `config.json`.

It's possible to customize the icons in the GH Alerts plugin itself:

https://github.com/antfu/markdown-it-github-alerts/blob/f2bcca68854891e13e7df77420734ac8cce6ebae/src/index.ts#L16-L22

But there was a problem where setting any custom icons would discard the defaults for the 5 default GH Alert icons (todo: re-confirm, was this the case?).

### DIFF
--- a/static/markdown.css
+++ b/static/markdown.css
@@ -149,6 +149,14 @@ blockquote {
     vertical-align: text-bottom;
     fill: currentColor;
 }
+/* default style for custom markers (Obsidian Callout style) */
+.markdown-alert {
+    border-left: .25rem solid var(--alert-custom-default);
+}
+.markdown-alert .markdown-alert-title {
+    color: var(--alert-custom-default);
+}
+/* default styles for GitHub style markers */
 .markdown-alert-note {
     border-left: .25rem solid var(--alert-note);
 }
--- a/static/markdown.css
+++ b/static/markdown.css
@@ -149,12 +149,12 @@ blockquote {
     vertical-align: text-bottom;
     fill: currentColor;
 }
-/* default style for custom markers (Obsidian Callout style) */
+/* default style for unconfigured custom markers (Obsidian Callout style) */
 .markdown-alert {
-    border-left: .25rem solid var(--alert-custom-default);
+    border-left: .25rem solid var(--alert-note);
 }
 .markdown-alert .markdown-alert-title {
-    color: var(--alert-custom-default);
+    color: var(--alert-note);
 }
 /* default styles for GitHub style markers */
 .markdown-alert-note {
--- a/src/parser/markdown.ts
+++ b/src/parser/markdown.ts
@@ -71,8 +71,10 @@ mdit.use(anchor, {
     }),
 });
 mdit.use(copycode);
+mdit.use(githubAlerts, {
+    markers: '*',
+});
 mdit.use(graphviz);
-mdit.use(githubAlerts);
 mdit.use(mermaid);
 mdit.use(wikiLinks);
 
--- a/static/colors.css
+++ b/static/colors.css
@@ -47,6 +47,8 @@
     --alert-warning: #ffaf00;
     --alert-caution: #ff5f5f;
 
+    --alert-custom-default: #c0c0c0;
+
     --ipynb-bg-error: rgba(255, 0, 0, 0.1);
     --ipynb-ansi-black: #616161;
     --ipynb-ansi-brblack: #8e8e8e;
@@ -118,6 +120,8 @@
         --alert-important: #8250df;
         --alert-warning: #bf8700;
         --alert-caution: #cf222e;
+
+        --alert-custom-default: #404040;
 
         --ipynb-bg-error: rgba(255, 0, 0, 0.1);
         --ipynb-ansi-black: #3e424d;
--- a/static/colors.css
+++ b/static/colors.css
@@ -47,8 +47,6 @@
     --alert-warning: #ffaf00;
     --alert-caution: #ff5f5f;
 
-    --alert-custom-default: #c0c0c0;
-
     --ipynb-bg-error: rgba(255, 0, 0, 0.1);
     --ipynb-ansi-black: #616161;
     --ipynb-ansi-brblack: #8e8e8e;
@@ -120,8 +118,6 @@
         --alert-important: #8250df;
         --alert-warning: #bf8700;
         --alert-caution: #cf222e;
-
-        --alert-custom-default: #404040;
 
         --ipynb-bg-error: rgba(255, 0, 0, 0.1);
         --ipynb-ansi-black: #3e424d;
--- a/tests/rendering/markdown-additional.md
+++ b/tests/rendering/markdown-additional.md
@@ -52,11 +52,13 @@ While not a markdown syntax, this has a default style:
 
 Press <kbd>Ctrl</kbd> + <kbd>C</kbd> to copy, and <kbd>Ctrl</kbd> + <kbd>V</kbd> to paste!
 
-## Custom attributes
+## Custom attribute
 
 This paragraph has a red background color.{style=background-color:red}
 
-## Github alert blockquote
+## GitHub style alert
+
+### The 5 default GitHub style alerts
 
 > [!NOTE]  
 > Something to take into account
@@ -73,7 +75,18 @@ This paragraph has a red background color.{style=background-color:red}
 > [!CAUTION]
 > Do not do this and that!
 
-With a custom title:
+### With a custom title
 
 > [!NOTE] Foo bar
-> Hello
+> 'Note' with a custom title
+
+### Using custom markers ([Obsidian Callout](https://help.obsidian.md/Editing+and+formatting/Callouts) style)
+
+> [!CUSTOM]
+> Something more special
+
+> [!fOoBaR]
+> The marker is case-insensitive and turns into Title Case
+
+> [!CUSTOM] paY aTtEntiOn
+> You can use a custom title with a custom marker as well
--- a/tests/rendering/markdown-additional.md
+++ b/tests/rendering/markdown-additional.md
@@ -77,13 +77,15 @@ This paragraph has a red background color.{style=background-color:red}
 
 ### With a custom title
 
-> [!NOTE] Foo bar
-> 'Note' with a custom title
+> [!TIP] Foo bar
+> 'Tip' with a custom title
 
 ### Using custom markers ([Obsidian Callout](https://help.obsidian.md/Editing+and+formatting/Callouts) style)
 
 > [!CUSTOM]
-> Something more special
+> Set custom icon and color for any marker
+>
+> If unconfigured, it's styled like 'Note'
 
 > [!fOoBaR]
 > The marker is case-insensitive and turns into Title Case


tuurep:
I wrote in a commit message that live-editing an incomplete alert marker causes page jitter

I was mistaken it's actually the Mermaid graph above the alerts in our testing doc (and we've talked about that a lot)

So no issue on alerts' side :) Behaves very well live-editing.

package.json
```diff
@@ -7,7 +7,7 @@
     "viv": "VIV_PORT=3000 node --loader ts-node/esm src/app.ts",
     "lint": "eslint src static",
     "lint-markdown": "markdownlint-cli2 --config .github/.markdownlint-cli2.yaml",
-    "test": "node --loader ts-node/esm tests/unit/cli.ts",
+    "test": "node --loader ts-node/esm --test 'tests/unit/*'",
```
tuurep, line 10: CI fails with this:

```
Run yarn test
yarn run v1.22.22
$ node --loader ts-node/esm --test 'tests/unit/*'
Could not find '/home/runner/work/Vivify/Vivify/tests/unit/*'
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Error: Process completed with exit code 1.
```

tuurep: Single quotes to double quotes didn't help

tuurep: Giving two explicit paths like `--test tests/unit/cli.ts --tests/unit/alerts.ts` at least reveals a bit about the problem:

```
Could not find '/home/runner/work/Vivify/Vivify/--test'
```

tuurep: Ok it can now run the tests but fails on one of them, let's see.

docs/customization.md
```diff
@@ -74,6 +74,96 @@ keys:
...
+
+      1. Name of an [octicon](https://primer.style/octicons/) (in kebab-case)
+      2. Path to an svg file, either relative to Vivify's config directory, or
+      an absolute path. Tilde (`~`) can be used in an absolute path.
+      3. A raw svg tag as a string. Note: must escape double quotes or use
+      single quotes inside the string.
+
+    > [!WARNING]
+    > We're considering switching from octicons to lucide as the main icon
+    > provider, so option 1 is subject to change.
```
tuurep, line 112: Double checked if the markdown renders on GitHub because it has weird syntax hl on the diff

Ironically, GitHub can't render this GitHub Alert, but Vivify can :smile:

tests/unit/alerts.ts
```diff
@@ -0,0 +1,66 @@
...
+        assertFallback(resolveIcon('./alert-icons/gear.txt'));
+        assertFallback(resolveIcon('../../svgs/dog.ico'));
+    });
+
+    test('wrong tag', () => {
+        // Considered to be (invalid) octicon names
+        assertFallback(resolveIcon("<img src='flower.png' alt='Alert icon hehe'>"));
+        assertFallback(resolveIcon('<script>document.createElement("svg");</script>'));
+    });
+});
```
jannis-baum, line 66: Is it intentional that we're only testing fallback cases and none of the ones that work?

src/parser/alerts.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+const MarkdownItAlerts = (md: MarkdownIt) => {
+    // Allow multi word alphanumeric markers (includes underscore)
+    // Additionally allow dashes in marker
+    const markerRE = '[\\w- ]+';
+
+    // Match marker case insensitively
+    // Note: config icons and titles keys must be fully lowercase
+    const RE = new RegExp(`^\\\\?\\[\\!(${markerRE})\\]([^\\n\\r]*)`, 'i');
```
jannis-baum, line 91: This can be simplified a lot if we use [Javascript's literal regex syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions#creating_a_regular_expression). But let's first talk about what the expression looks like as it is:

```
^\\?\[!([\w- ]+)\]([^\n\r]*)
```

What I find strange about this is that it looks for an optional backslash in the beginning, i.e. you can write `\[!note] title`, is that part of the spec? If not or we don't know I'd drop it.

Dropping it & using the literal syntax we arrive to

```suggestion
    // Match marker case-insensitively
    // Note: config icons and titles keys must be fully lowercase
    const RE = /^\[!([\w\- ]+)\]([^\n\r]*)/i;
```

Beyond this I also think the variable name `RE` is very unusual haha so some camelCased proper name would be better.

src/parser/alerts.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    md.core.ruler.after('block', 'alerts', (state) => {
+        const tokens = state.tokens;
+        for (let i = 0; i < tokens.length; i++) {
+            if (tokens[i].type === 'blockquote_open') {
```
jannis-baum, line 96: Prefer negative case & continue, i.e.

```suggestion
            if (tokens[i].type !== 'blockquote_open') continue;
```

So that we can have the rest be one indent level less

jannis-baum (changes requested):
Looks great overall & works, thanks a lot for the commitment in making this good!

I left some questions and also some improvements for the code you probably got from that plugin

tuurep:
Took a glance, but getting back after tuesday

The case-insensitive config-keys is the most important thing, left a reply on that one in advance

docs/customization.md
```diff
@@ -74,6 +74,96 @@ keys:
...
+    "fallbackIcon": "alert"
+  }
+  ```
+
+  In more detail:
+
+  - **`"icons"`**
+
+    **Key:** Any custom marker name. Can also be a multi-word string. Has to be
+    fully lowercased.
```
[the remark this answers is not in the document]
jannis-baum: Hm you could implement

```typescript
function lowerCaseKeys(obj: Record<string, string> | undefined) {
    if (obj === undefined) return obj;
    return Object.keys(obj).reduce(
        (accumulated, key) => {
            accumulated[key.toLowerCase()] = obj[key];
            return accumulated;
        },
        {} as Record<string, string>,
    );
}
```

and then instead of `config.alertsOptions?.icons` you use `lowerCaseKeys(config.alertsOptions?.icons)`, same with the titles. Can't think of a cleaner way ^^

src/config.ts
```diff
@@ -18,6 +18,7 @@ type Config = {
     /* eslint-disable @typescript-eslint/no-explicit-any */
     katexOptions?: any;
     tocOptions?: any;
+    alertsOptions?: any;
     /* eslint-enable @typescript-eslint/no-explicit-any */
```
jannis-baum, line 22: Just saw that this wasn't typed while looking at the lower case thing, we should definitely fix that. From the `customization.md` I read:

```suggestion
    /* eslint-enable @typescript-eslint/no-explicit-any */
    alertsOptions?: {
        icons?: Record<string, string>;
        titles?: Record<string, string>;
        fallbackIcon?: string;
    };
```

Is that right?

(This differs from `katexOptions` and `tocOptions` because those are plugins that don't expose the correct types, that's why we have `any` there. For our own code we should type properly)

static/markdown.css
```diff
@@ -125,60 +125,55 @@ blockquote {
...
 }
 
-/* GitHub-style alerts */
-.markdown-alert {
+/* Alerts: */
+/* (also known as Callouts or Admonitions) */
+.alert {
+    /* Default color variable: overridden per marker */
+    /* Falls back to --fallback-alert for unconfigured markers */
+    --color: var(--fallback-alert);
```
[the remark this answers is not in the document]
tuurep: Yeah this ended up surprisingly clean :+1:

src/parser/alerts.ts
tuurep: Note: this part comes straight from original plugin and I didn't look at it, but yes let's make it as good as possible

src/config.ts
tuurep: Thanks, was due to inexperience with TS :)

tuurep: I'll do this along with naming change `=> alertOptions`

src/parser/alerts.ts
tuurep: > What I find strange about this is that it looks for an optional backslash in the beginning, i.e. you can write `\[!note]` title, is that part of the spec? If not or we don't know I'd drop it.

Yes I don't understand the point to this, let's try removing it

src/parser/alerts.ts
tuurep: Addressed by [cdafc3d](https://github.com/jannis-baum/Vivify/pull/233/commits/cdafc3d4ba6003366a628ad13de652583e83e559)

docs/customization.md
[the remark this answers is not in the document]
tuurep: I think what I did in c4ff883 was a very simple way to do this:

```ts
const resolvedIcons: Record<string, string> = {};
for (const marker in icons) {
    resolvedIcons[marker.toLowerCase()] = resolveIcon(icons[marker]);
}

const titles: Record<string, string> = {};
for (const marker in config.alertOptions?.titles ?? {}) {
    titles[marker.toLowerCase()] = config.alertOptions!.titles![marker];
}
```

Is this ok?

jannis-baum (commented):
Good job with the updates! I think you still wanted to move the detailed explanation of the alerts from the customization to its own document right? Anything else missing aside from that?

tuurep:
Yeah writing a guide-like doc with some screenshot examples is the main thing

Was wondering if there's any chance at making this bit more clean: https://github.com/jannis-baum/Vivify/pull/233#discussion_r2317535925

But right now, it works.

Do you think the unit tests look good enough?

tuurep:
Didn't exactly mean to push that WIP docs commit, I was doing a rebase here

Going to finish the docs by the end of tomorrow

src/parser/alerts.ts
```diff
@@ -81,40 +81,57 @@ function capitalize(str: string): string {
...
+            // Get the first inline token, consists of:
+            // 1. title line e.g. [!marker] Optional title
+            // 2. the first paragraph in the alert body
+            let firstBlock;
+            for (let j = start; j <= end; j++) {
+                if (tokens[j].type === 'inline') {
+                    firstBlock = tokens[j];
+                    break;
+                }
             }
```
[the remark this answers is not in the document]
jannis-baum: Hm I looked at that by

```typescript
            // Get the first inline token, consists of:
            // 1. title line e.g. [!marker] Optional title
            // 2. the first paragraph in the alert body
            let firstBlock;
            for (let j = start; j <= end; j++) {
                console.log(j - start + 1);
                console.log(tokens[j]);
                if (tokens[j].type === 'inline') {
                    console.log('XXXXXXXXXXXXXX');
                    firstBlock = tokens[j];
                    for (; j <= end; j++) {
                        console.log(j - start + 1);
                        console.log(tokens[j]);
                    }
                    break;
                }
            }
```

And I don't think we can influence it. Seems to come down to whether or not people add an empty (quote) line between the title line and actual first paragraph, e.g. in

```markdown
> [!note]
> hehe
```

the alert title and alert first paragraph are just one paragraph in vanilla markdown so markdown-it treats it as a single token. This

```markdown
> [!note]
>
> hehe
```

on the other hand are two vanilla paragraphs so there you do get two tokens.

jannis-baum:
> Was wondering if there's any chance at making this bit more clean: [#233 (comment)](https://github.com/jannis-baum/Vivify/pull/233#discussion_r2317535925)

I replied there!

> Do you think the unit tests look good enough?

Yes looks good to me! I realized the tests weren't being linted so I pushed a fix for that (including making the linting pass by auto-fixing your file), hope that's okay.

src/parser/alerts.ts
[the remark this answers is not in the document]
tuurep: Thanks,

In that case it's probably fine the way it is now

tuurep:
@jannis-baum basically finished!

About the screenshots: this doc ended up quite long and challenging to write in a way that the info is clear..

With that, I might be reconsidering the screenshots, especially as they require extra maintenance with the assets repo, don't look good in dark mode, etc.

What do you think, better with or without image examples?

For instance here are the ones from the example section:

image

jannis-baum:
Good job on the documentation, I think it's nice and understandable! See the small comments above.

About the screenshot, I agree that it should do fine without them.

docs/alerts-guide.md
```diff
@@ -0,0 +1,214 @@
...
+{
+  "alertOptions": {
+    "fallbackIcon": "alert"
+  }
+}
+```
+
+Note: `"fallbackIcon"` is a separate option outside of the `"icons"` table, but
+accepts the same type of icon value as the [icon
+customization](#configuring-alert-icons).
```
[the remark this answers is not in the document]
tuurep: Sure, would have been a little confusing with example alerts in the doc but in the end, the markdown lint didn't let me list example alerts the way I wanted to lol

README.md
```diff
@@ -42,15 +42,18 @@ issue](https://github.com/jannis-baum/vivify/issues/new/choose) or
...
   as well as absolute file links
 - [add styles, classes, ids or other attributes directly from
   Markdown](https://github.com/arve0/markdown-it-attrs?tab=readme-ov-file#examples)
 - table of contents with `[[toc]]`
 - `<kbd>` tags, e.g. to style keyboard shortcuts
+- [GitHub alert
+  blocks](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)
+  - with syntax extension similar to [Obsidian callouts](https://help.obsidian.md/callouts)
+  - configurable [custom alerts](docs/alerts-guide.md), with defaults like
+    GitHub
```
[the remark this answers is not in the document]
tuurep: Sure, moved and reworded,

didn't directly mention the 'superset' concept, but I think the message comes across?

tuurep:
Ok I force-amended to the last commit adding your suggestions with my best judgement, could you check one more time?

PR #238, opened 2025-08-30
Issue/234 fenced code blocks render internal html

### LINKED ISSUE #234: Fenced code blocks render internal HTML
### Description

Fenced code blocks that contain html, xml or similar style syntax and don't explicitly set a syntax highlighting language will try to render as html.

### To Reproduce

Based on the extended syntax test (just without the language set)

````
```
<html>
  <head>
  </head>
  <body>
    <p style="font-size: 1000px">This is some really huge text!</p>
  </body>
</html>
```
````

#### Expected result

Image

#### Actual result

Image

#### Additional

Expected result tested and is consistent with GitHub, VSCode, Obsidian. Work around is to explicitly set the lang type to `text` if you want no highlighting, or `html` / `xml` for formatting with highlighting.

### Context

viv version: v0.8.2 (on Fedora 42)

### PR DESCRIPTION
**Pull request by Tweekism**
Fixes #234

### DIFF
--- a/src/parser/highlight.ts
+++ b/src/parser/highlight.ts
@@ -2,15 +2,16 @@ import hljs from 'highlight.js';
 
 export default function highlight(str: string, lang: string) {
     let content = str;
-    if (lang) {
-        if (hljs.getLanguage(lang)) {
-            try {
-                content = hljs.highlight(content, {
-                    language: lang,
-                    ignoreIllegals: true,
-                }).value;
-            } catch {}
-        }
+    if (!lang) {
+        lang = 'text';
+    }
+    if (hljs.getLanguage(lang)) {
+        try {
+            content = hljs.highlight(content, {
+                language: lang,
+                ignoreIllegals: true,
+            }).value;
+        } catch {}
     }
     return `<pre class="language-${lang}"><code>${content}</code></pre>`;
 }
--- a/src/parser/highlight.ts
+++ b/src/parser/highlight.ts
@@ -2,16 +2,13 @@ import hljs from 'highlight.js';
 
 export default function highlight(str: string, lang: string) {
     let content = str;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/rendering/markdown-extended.md [hunk 1]
```diff
@@ -19,6 +19,17 @@ https://www.markdownguide.org/extended-syntax/
 
 ## Fenced code block
 
+A normal fenced code block with no syntax highligting:
+
+```
+# PLain text
```
jannis-baum, line 25: typos🙈

``````suggestion
A normal fenced code block with no syntax highlighting:

```
# Plain text
``````

File: src/parser/highlight.ts [hunk 2]
```diff
@@ -2,15 +2,16 @@ import hljs from 'highlight.js';
...
+        lang = 'text';
+    }
+    if (hljs.getLanguage(lang)) {
+        try {
+            content = hljs.highlight(content, {
+                language: lang,
+                ignoreIllegals: true,
+            }).value;
+        } catch {}
     }
```
jannis-baum, line 15: If we do it like this there's still the edge case where `hljs.getLanguage` returns `undefined` for unknown languages, e.g.

``````markdown
```hehe
<h1>this will render as html</h1>
```
``````

so what we could also do instead would be:

```suggestion
    if (!hljs.getLanguage(lang)) lang = 'text';
    try {
        content = hljs.highlight(content, {
            language: lang,
            ignoreIllegals: true,
        }).value;
    } catch {}

```

Tweekism: Yeah that's a good idea actually. I'll wrap this up in a nice commit now and test.

Does this mean both of the original nested `if` blocks were kind of useless? Would not the result been with just the try .. catch on its own?

jannis-baum (approved):

PR #240, opened 2025-08-31
Links to local html files & `From...` mime type fix

### LINKED ISSUE #235: Markdown considered filetype "Mailbox text" if it starts with `From ...`

### Description

If a markdown file starts with: `"From ..."`, it's considered to be filetype `"Mailbox text"`.

For me, this causes Firefox to open a `"Save file as..."` dialog instead of rendering as markdown (though the exact behavior depends on a Firefox setting).

### To Reproduce

1. Create file `test.md`:

   ```md
   From xahlee.info:

   > Emacs is possible because of the embedded programing language emacs lisp. Emacs can be
   > considered as an interactive user interface to emacs lisp. 

   ```

2. Attempt to view it with `$ viv test.md`

Minimally, just needs the `"From"` at start followed by anything, e.g. this file is enough to reproduce:

```md
From John
```

### Context

- Arch Linux

### LINKED ISSUE #236: Links to local html files
Currently vivify will correctly handle URLs pointing to websites - sending you to the website in the browser you are using to view vivify's output.

Sometimes I have local html files that I would like to be handled the same way - so I can link to local html files in my markdown wiki and easily browse to them via the link when using Vivify.

Currently, Vivify instead tries to preview the html by showing the raw html within vivify.  Would it be possible to change this behaviour so that if I click on a link to a local html file, the local html file is opened in my browser as if it were a website?

### LINKED ISSUE #248: Codeblock looks 1 line too long in code viewing

### Description

Image

Ref: https://github.com/jannis-baum/Vivify/pull/240#issuecomment-3268538586

> I think that's because of the trailing space in files.

### To Reproduce

`viv module.mjs` (or any file that should render as plaintext/code)

jannis-baum:
Close #236
Close #235 
Close #248

### DIFF
--- a/tests/rendering/html/index.html
+++ b/tests/rendering/html/index.html
@@ -6,7 +6,7 @@
     <body>
         <h1>HTML rendering test</h1>
         <p>This should only render if Vivify is configured to render HTML.</p>
-        <p>If styling works, these paragaphs should be green.</p>
+        <p>If styling works, these paragraphs should be green.</p>
         <p>If scripting works, we should see "Script works" and "Module works" paragraphs below.</p>
         <script type="text/javascript" src="script.js"></script>
     </body>


jannis-baum:
Hey @tuurep this is actually *very* close to auto-refreshing :D I already implemented injecting the Vivify client code into the HTML files so the only unresolved thing towards auto-refreshing is that the `UPDATE` client function right now tries to replace the content in the element with ID `body-content` that of course doesn't exist. If we figure out that last part you have your auto-refresh as well🥳 But let's not do that in this PR

tuurep (approved):
This is awesome, I suppose it's already enough that instead of `$ firefox <file>.html` I can always to `viv <file>.html` instead

Tested:
- `viv` on an existing HTML file :heavy_check_mark: 
- whether the file where I discovered the `application/mbox` thing works :heavy_check_mark:

`viv` on the test file you added:

image

I'll definitely always have `"renderHTML": true`, but hard to say if the overwhelming majority would prefer that.

tuurep:
Oh yeah, I didn't quite understand why CSS, JS and TS files would then render like this for example:

image

But it looked very intentional in the code so was there some big reason it has to be like it?

jannis-baum:
> Oh yeah, I didn't quite understand why CSS, JS and TS files would then render like this for example:
> 
> ...
> 
> But it looked very intentional in the code so was there some big reason it has to be like it?

Yeah that's the trade-off :( The thing is that when you reference a local CSS or JS from HTML it still has to go through Vivify because the browser doesn't let you switch hosts (e.g. from Vivify to `file://`) for security reasons.

That means Vivify has to serve everything we need for HTML as plain text so the browser understands it. I can't think of any way we could distinguish if the client is the browser requesting the files to render HTML or if the actual user is looking at them. Actually now that I'm writing this, maybe there's something in the HTTP request header👀 should check that before we merge, that would be awesome

tuurep:
Testing on firefox plus initial impressions:

- [ ] Firefox
    - [x] png
    - [x] jpg
    - [x] svg
    - [x] avif
    - [x] html
    - [x] css
    - [x] js
    - [ ] ts

For some reason JS renders like intended, but TS still opens this way:

image

Could it be a simple mistake?

Everything else works fantastic and image browsing was cool to try.

Something I noticed about the codeblock when viewing code on Vivify:

image

But I confirmed that's always been the case and not specific to JS :smile: 

HTML feels different to other filetypes in that it can't be navigated with a back button, but I don't necessarily think it should (but actually I always hide the back button in my own configs).

I'm going to try separately with another browser.

tuurep:
The situation is exactly the same on Vivaldi (Chromium-based) with all of the above.

Btw another thing I noticed but a little off-topic, is that we can't follow symbolic links in Vivify's file browser:

image

Not sure if that's complicated to fix; if not, we should.

tuurep:
Oh, this should be relevant: even with `"renderHTML": false`, TS opens the same way.

src/parser/parser.ts
```diff
@@ -28,33 +28,44 @@ function wrap(contentType: string, content: string, linkPath?: string): string {
...
+            renderMarkdown(`${pathHeading(path)}\n\n![${path}](./${pbasename(path)})\n`),
+            pdirname(path),
+        );
     }
-}
 
-export const shouldRender = (mime: string): boolean =>
-    mime.startsWith('text/') || mime === 'application/json' || mime === 'inode/x-empty';
+    // render text files
+    if (mime.startsWith('text/') || mime === 'application/json' || mime === 'inode/x-empty') {
```
tuurep, line 47:
```suggestion
    if (
        mime.startsWith('text/') ||
        mime === 'application/javascript' ||
        mime === 'application/json' ||
        mime === 'inode/x-empty'
    ) {
```

Looks like this works (with added `application/javascript`), just confusing that viewing JS worked without this.

src/utils/path.ts
```diff
@@ -1,12 +1,22 @@
...
+    const ext = extname(path).slice(1);
+    if (config.mdExtensions.includes(ext)) {
+        return 'text/plain';
+    } else if (ext == 'css') {
+        return 'text/css';
+    } else if (ext == 'html') {
+        return 'text/html';
+    } else if (['js', 'mjs'].includes(ext)) {
+        return 'text/javascript';
+    }
```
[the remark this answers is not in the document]
tuurep: Consider whether we want to add also: `tsx`, `jsx`, `cjs`

(I don't know if more exist)

jannis-baum:
By the way @tuurep while I think truly live previewing HTML is not possible without horrible hacks due to browser security, something that would be doable now with relatively low effort is making it sync when the file is written to. Would that something you would like?

jannis-baum:
Ok so I guess we have to decide between Node mime or adding the application/json check?

Yes. And either option comes with side effects (application/json check might not be the only thing we'd have to add vs. not working for extentionless files)

I don't really know. But actually, does extensionless mimetype checking work currently? Here's an example of me viewing my .zshrc (no syntax hl):

Yes it does, in your example the `.zshrc` was correctly identified as text. The syntax highlighting doesn't use the MIME system, it just goes off of file extensions and I guess `.zshrc` is not a file extension the syntax highlighter knows

src/routes/viewer.ts
```diff
@@ -42,14 +51,37 @@ router.get(/.*/, async (req: Request, res: Response) => {
...
+                        // If there's one closing tag for the body we add the
+                        // Vivify client script to the end of the body. If not
+                        // then we try to just add it to the end of the
+                        // document.
+                        const bodyEndCount = (html.match(/<\/body>/g) || []).length;
+                        if (bodyEndCount === 1) {
+                            html = html.replace('</body>', `${vivClient(req)}</body>`);
+                        } else {
+                            html = html + vivClient;
+                        }
```
tuurep, line 78:
```suggestion
                        } else {
                            html = html + vivClient(req);
                        }
```

Fixes a case where

```
<html>
    <h1>Hello</h1>
</html>
```

would render as:

image

Any problems with that? It was 100% a ChatGPT suggestion,

tuurep:
Is it possible to control the page title when the HTML is rendered?

Would be nice if it could be other than:

`http://localhost:31622/viewer/~/test.html`

image

PR #253, opened 2025-09-10
Improve symlinks in directory viewer

Tweekism:
Hi team,

A work-in-progress draft PR for the symbolic link implementation in the directory view renderer.

I haven't checked yet, but there should be one line that fails Prettier linting, that has some hard coded color values that will need to be moved to the colors.css anyway (once we pick nicer colors). 

This is all i have for today, @tuurep if you want to play with it you are welcome 😋

### DIFF
--- a/src/parser/parser.ts
+++ b/src/parser/parser.ts
@@ -1,4 +1,4 @@
-import { Dirent } from 'fs';
+import { Dirent, readlinkSync, realpathSync } from 'fs';
 import { homedir } from 'os';
 import { join as pjoin, dirname as pdirname, basename as pbasename } from 'path';
 import { pathToURL } from '../utils/path.js';
@@ -11,6 +11,7 @@ import octicons from '@primer/octicons';
 const dirIcon = octicons['file-directory-fill'].toSVG({ class: 'icon-directory' });
 const fileIcon = octicons['file'].toSVG({ class: 'icon-file' });
 const backIcon = octicons['chevron-left'].toSVG({ class: 'icon-back' });
+const linkIcon = octicons['file-symlink-file'].toSVG({ class: 'icon-symlink' });
 
 export type Renderer = (content: string) => string;
 
@@ -57,12 +58,29 @@ export function renderTextFile(content: string, path: string): string {
     return wrap(contentType, render(content), pdirname(path));
 }
 
-const dirListItem = (item: Dirent, path: string) =>
-    `<li class="dir-list-${item.isDirectory() ? 'directory' : 'file'}" name="${item.name}">
-        <a href="${pathToURL(pjoin(path, item.name))}">
-            ${item.isDirectory() ? dirIcon : fileIcon}${item.name}
+function dirListItem(item: Dirent, path: string): string {
+    let itemClass = '';
+    let icon = '';
+    let fullPath = '';
+    if (item.isSymbolicLink()) {
+        itemClass = "dir-list-symboliclink";
+        icon = linkIcon;
+        fullPath = pjoin(path, readlinkSync(pjoin(path, item.name)));
+    } else if (item.isDirectory()) {
+        itemClass = "dir-list-directory";
+        icon = dirIcon;
+        fullPath = pjoin(path, item.name);
+    } else {
+        itemClass = "dir-list-file";
+        icon = fileIcon;
+        fullPath = pjoin(path, item.name);
+    }
+    return `<li class="${itemClass}" name="${item.name}">
+        <a href="${pathToURL(fullPath)}">
+            ${icon}${item.name}
         </a>
     </li>`;
+}
 
 function dirUpItem(path: string): string {
     if (pbasename(path) == '') {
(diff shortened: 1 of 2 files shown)


tuurep:
After icon size & spacing tweaks:

image

On hover, destination part no longer gets a *blue* underline:

image

tuurep:
On dark theme:

image

IMO this looks great, I'd be pleased at this point with the style and will move on to thinking about the 404 page.

tuurep:
How it looks in a real world scenario, exploring some system dirs:

image

tuurep:
We found a bug: absolute paths don't work when clicked on because the path joins wrong (simple to fix)

src/parser/parser.ts
```diff
@@ -73,12 +75,26 @@ export function renderBody(
...
-        <a href="${pathToURL(pjoin(path, item.name))}">
-            ${item.isDirectory() ? dirIcon : fileIcon}${item.name}
-        </a>
-    </li>`;
+function dirListItem(item: Dirent, path: string): string {
+    if (item.isSymbolicLink()) {
+        const targetPath = readlinkSync(pjoin(path, item.name));
+        const isValid = existsSync(pjoin(path, targetPath));
+        return `<li class="dir-list-symlink" name="${item.name}">
+                    <a href="${isValid ? pathToURL(pjoin(path, targetPath)) : ''}">
```
Tweekism, line 83: This one, I suck at using the github code reference thing...

tuurep:
Thanks! Tested that it works,

found a new problem unrelated to that: when a symlink points to another symlink, (unsurprisingly) can't resolve:

image

I wouldn't have thought to make such a symlink, ~~thanks Steam~~

tuurep:
1. Attempt at fixing recursive symlinks without breaking anything else
2. Rebased and resolved conflicts (naming in `backIcon` resulted in conflicts)

src/parser/parser.ts
```diff
@@ -119,3 +153,7 @@ export function renderDirectory(path: string): string {
         ),
     );
 }
+
+export function renderErrorPage(errorMessage: string): string {
+    return renderMarkdown(`# Something went wrong\n\n\`\`\`${errorMessage}\`\`\``);
+}
```
tuurep: Is the "Something went wrong" really a good header,

It could also be the standard:

- 404 Not Found
- 500 Internal Server Error

?

jannis-baum: I'd say it's good enough because we don't have proper status codes for a lot of cases so I wouldn't display them explicitly. But if you think that's better I'm also good with it so feel free to decide :)

src/routes/viewer.ts
```diff
@@ -89,41 +122,16 @@ router.get(/.*/, async (req: Request, res: Response) => {
                     return;
                 }
             }
-        } catch (error) {
-            res.status(500).send(String(error));
+        } catch (error: unknown) {
+            const e = error as Error & { code?: string };
+            body = renderErrorPage(e.message);
+            const statusCode = e.code === 'ENOENT' ? 404 : 500;
+            res.status(statusCode).send(renderFullPage(title, body, req));
```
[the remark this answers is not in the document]
jannis-baum: I think it's fine since this is not a "proper API" or anything where correct/elaborate status codes are super important

src/parser/parser.ts
jannis-baum: Yes that's good

tuurep:
Right, that makes sense!

tuurep:
@jannis-baum Finished:

image

The colors are a little bit confusing, let me know what you think:

- GitHub uses directory color for all symlink icons (symlink-file as well)
- We use link color for symlink-file, and dir color for symlink-dir
    - The fact that our dark theme's dir color isn't blue may look confusing but we had plans to change the default to GitHub dark anyways

The most difficult part was to sort dir-pointing symlinks with the regular dirs, resulting in the most refactor: [50f9f87](https://github.com/jannis-baum/Vivify/pull/253/commits/50f9f87dc8c77ab203ba9bdb7675cf509cbd85bc)

Sorting looks solid, but let me know if the code is ok.

tuurep:
Other color compromises:

### 1. All symlink icons as dir color (exactly like GitHub):

```diff
 .icon-symlink-directory { fill: var(--icon-directory); }
-.icon-symlink-file { fill: currentColor; }
+.icon-symlink-file { fill: var(--icon-directory); }
```

image

### 2. All symlink icons as link color

```diff
-.icon-symlink-directory { fill: var(--icon-directory); }
+.icon-symlink-directory { fill: currentColor; }
 .icon-symlink-file { fill: currentColor; }
```

image

### 3. Symlink-dir as dir color, symlink-file as file color

```diff
 .icon-symlink-directory { fill: var(--icon-directory); }
-.icon-symlink-file { fill: currentColor; }
+.icon-symlink-file { fill: var(--icon-file); }
```

image

### Opinions

1. This could actually work and would just be exactly what GitHub does
2. Looks terrible
3. Looks ok and makes a lot of sense

tuurep:
If we want to go for option 3. and ignore what GitHub decides to do, we could also consider making the label look like just any other entry:

image

On hover:

light-hover

image

jannis-baum (approved):
Nice looks good, great job! Feel free to merge :)

PR #254, opened 2025-09-12
Auto-reload HTML

### LINKED ISSUE #251: Auto-reload HTML
We have talked about this a lot, e.g. in #240, #236, #84. Fully live-reloading HTML as we do it with Markdown is not feasible as far as I can tell since it would require replacing the whole document from client-side JS which is not allowed due to security restrictions. There is probably a super hacky way to do it but it would be complicated, prone to errors, and require parsing potentially super-invalid HTML, so I'd prefer not to.

After #240 is merged, making HTML reload on save however will be relatively straightforward to implement, that's what this issue is (*auto* vs. *live* reloading). I see two possible paths for this:

1. We keep the plugin pattern and add a `BufWritePost` autocmd to the plugin that tells Vivify to hard-reload HTML clients.
2. We implement watching files from Vivify and always hard-reload all clients that don't have live content when they are written to. Would make Vivify slightly more complex but keep functionality in the plugin to a minimum and you would get auto reloading on all files even without needing an editor plugin.

@tuurep what do you think about this?

jannis-baum:
Close #251

This was actually super small and simple now after #240 ^^

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,9 @@ features!
 - view image files
 - view & navigate directories using "back to parent directory" button at the
   top-left of the file viewer
-- easy to integrate with any editor for live synchronization (see [editor
-  support](#editor-support))
+- automatic reloads when files change
+- easy to integrate with any editor for fully live synchronization without
+  saving the file (see [editor support](#editor-support))
 - Vivify server starts lazily and automatically shuts down when no more viewers
   are connected
 - various [customization options](docs/customization.md)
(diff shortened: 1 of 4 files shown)


tuurep:
Yeah I can't find a way to break this and it's really solid! Makes even surprising things possible, like previewing changes to an svg:

https://github.com/user-attachments/assets/e2c13e86-4c32-4709-b215-e56a429ae05c

(although firefox is zoomed in as far as it can be)

I think some people could even find this enough to use with editors that are not (n)vim, for markdown too.

Fantastic feature, my favorite yet :smile:

PR #258, opened 2025-09-16
YAML formatter

### LINKED ISSUE #177: YAML formatter
Hi,

I feel like some ungrateful spammer, but here I am (sorry).

In the nutshell:
Would be possible to render `YAML Fromatter` differently or not at all?

I don't know how to explain so lets go for examples. I am using quite often in my md files at the very top of file yaml formatter

For example:
image

This is rendered like this:
image

So the `YAML` part in this case is treated as `H1.

But in this file:
image

it is rendered like this:
image

This case is a mess. It detects the list of tags and put every thing in it.

What I  would like to have? Something like this:
image

And by default it could be folded:
image

/Best regards
Adam

jannis-baum:
Close #177


jannis-baum:
Just realized that the OP of the issue wanted the front matter to just be displayed (and foldable) so I guess I'll have to add that as well. Would of course be possible right now with customization but super annoying because you only get the data (as JSON) and not the plain YAML (because that would make no sense in JS) so you'd have to convert it back lol

tuurep:
The collapse-button looks a little too high up

image

I just dealt with the same thing with the arrow icon on #253 so I can probably suggest a fix

tuurep:
I was able to align the collapse button with this:

```diff
diff --git a/src/parser/front-matter.ts b/src/parser/front-matter.ts
index 469c146..9de1ade 100644
--- a/src/parser/front-matter.ts
+++ b/src/parser/front-matter.ts
@@ -40,7 +40,7 @@ export default function parseFrontMatter(md: MarkdownIt) {
             </script>
             <div style="text-align: right">
                 <a href="" id="front-matter-button" onClick="frontMatterButtonClicked(); return false;">
-                    Front Matter <span id="front-matter-collapse-icon">${downIcon}</span>
+                    Front Matter ${downIcon}
                 </a>
             </div>
             <pre id="front-matter-display" style="display: none"><code>${yaml}</code></pre>
diff --git a/static/style.css b/static/style.css
index 5f18551..cbbc495 100644
--- a/static/style.css
+++ b/static/style.css
@@ -20,8 +20,8 @@ body {
 }
 .content-txt { max-width: 1200px; }

-/* parent directory link */
-#top-nav > a {
+/* buttons at the top */
+#top-nav-up, #front-matter-button {
     display: inline-flex;
     align-items: center;
 }
```

image

But would be nice to keep the `front-matter-collapse-icon` id for customization

tuurep:
Should Front Matter button be hidden on print?

`style.css`

```diff
 @media print {
-    #top-nav {
+    #top-nav, #front-matter-section {
         display: none;
     }
```

tuurep (changes requested):
TL;DR my requests:

- Consider if we can put the front matter button on the same level as the back button: https://github.com/jannis-baum/Vivify/pull/258#issuecomment-3304816942
- Vertically align collapse button: https://github.com/jannis-baum/Vivify/pull/258#issuecomment-3304888062
- Hide Front Matter button on print mode? https://github.com/jannis-baum/Vivify/pull/258#issuecomment-3305092946
- One typo in docs: https://github.com/jannis-baum/Vivify/pull/258#discussion_r2357067513

Tested the code enough to say that it works, though I'm very unfamiliar with the feature, I would say looks good to me.

jannis-baum:
Okay so I ended up implementing having the front matter button moved into the top nav with client-side JS on page load. I actually don't feel like it's too bad. Shouldn't really hurt customization because that just runs once and you can just customize from the state it ends up with after that has run.

tuurep:
Nice, looks great

Gave it another try to see how it looks when live editing:

https://github.com/user-attachments/assets/f0be1028-f7f1-4ab5-9774-455036fd2930

Pictured:

1. when creating the thing, wraps whole document in the front matter
    - not an issue really
2. jitter of the button position on live editing
3. unable to keep the fold open when editing
    - it's closed on the first refresh

What do you think about these? Could be the case that it's not important to see real-time changes to front matter's contents as well?

tuurep:
Errors seen on garbage YAML:

image

I tried to look for a `Failed to parse front matter JSON` but maybe there's a problem earlier than that :)

I guess most important is that it doesn't completely crash.

jannis-baum:
> Pictured in video:
> 
> 1. when creating the thing, wraps whole document in the front matter
>    * not an issue really
> 2. jitter of the button position on live editing
> 3. unable to keep the fold open when editing
>     * it's closed on the first refresh
>     * Could be the case that it's not important to see real-time changes to front matter's contents as well?
> 
> What do you think about these? 

(quote slightly adapted)

1. I agree it's not an issue, there's no way around that since that is the valid syntax ^^
2. I managed to fix this one by doing the "move button into nav area" server-side already. Much cleaner this way in a lot of ways. Thanks for pushing!
3. I agree that it's probably not important to see real-time updates there. Storing the state of the fold across reloads would be kind of annoying and not worth it I think [^1]

[^1]: We'd have to move the folding script into the `client.mjs` to store state there and then adapt the CSS `display` property back to the current state on each `UPDATE`. I kind of like that right now if you're not using front matter there won't be any trace of it in the client script so that would no longer work then.

Issue #250, opened 2025-09-10
Alerts: Don't render Obsidian fold character `-` or `+` as a title


tuurep:
Followup #233

Obsidian supports the following syntax to collapse (fold) the alert contents:

`> [!note]-`

In our current alerts implementation, this creates an alert with the *title* `-` (a dash)

This hurts viewing files written with Obsidian.

A likely solution would be: only parse the custom title outside the square brackets when there's a space between bracket and title. It's probably just surprising behavior anyway, that the title renders without the space.

This could be solved at the same time as #245, they concern editing the same regex.

PS. Vivify doesn't support the Obsidian folding syntax because I feel like that's a bigger rabbit hole (i.e. what else should be foldable) and there's no precedent for folding an alert styled in the GitHub way.

PR #270, opened 2025-10-17
UTF-8 support in alert marker names

### LINKED ISSUE #245: UTF-8 support in alert `[!marker names]`
Followup #233

This is something that is supported in Obsidian, but needs looking into in our implementation.

Be able to use marker name in a language that isn't English:

```md
> [!hyvää päivää]
```

```md
> [!你好]
```

But actually, we can't even use an apostrophe right now:

```md
> [!it's sad]
```

Needs:

1. modify the regex in `src/parser/alerts.ts`
    - `/^\[!([\w\- ]+)\]([^\n\r]*)/i`
2. make sure CSS classes can be used when they contain UTF-8:    
    - `.alert-hyvää-päivää`
    - `.alert-你好`
3. actually the apostrophe case is more complicated: would the single quote have to be escaped for the class name?
    - `.alert-it\'s-sad`

tuurep:
Resolves #245 and #250 

- [x] Ignore Obsidian fold character
- [x] Allow any characters inside `[!marker]`
    - [x] Edge cases with CSS-classnames: for example `"` used in marker name

### DIFF
--- a/src/parser/alerts.ts
+++ b/src/parser/alerts.ts
@@ -142,7 +142,11 @@ const MarkdownItAlerts = (md: MarkdownIt) => {
     });
     md.renderer.rules.alert_open = function (tokens, idx) {
         const { marker, title, icon, isFallback } = tokens[idx].meta;
-        const markerId = marker.replace(/\s+/g, '-');
+
+        const markerId = marker
+            .replace(/\s+/g, '-') // get rid of spaces in a CSS classname
+            .replace(/"/g, '&quot;'); // escape quotes so they don't break the HTML tag
+
         return `<div class="alert alert-${markerId} ${isFallback ? 'fallback-alert' : ''}">
                     <p class="alert-title">${icon}${title}</p>`;
     };


jannis-baum (commented):
Nice, cool that it was so easy! Could you add an example for this to the rendering test for markdown-additional?

tuurep:
Escaping is going to work in some way:

https://stackoverflow.com/questions/448981/which-characters-are-valid-in-css-class-names-selectors

But going to take some investigation about the exact set of characters that need escaping for a CSS class

To use in `styles.css`, I think there'll be these two ways to select them:

```css
.alert-\"hello-world\"
[class~='alert-"hello-world"']
```

So, looking promising.

tuurep:
Finished

Fortunately most of the complexity falls on correctly escaping in the selector in a `styles.css` and all we needed to add was to escape the quote so the HTML tag doesn't break.

Examples targeting the `"hello world"` marker:

```json
{
    "styles": "styles.css",
    "alertOptions": {
        "icons": {
            "\"hello world\"": "zap"
        }
    }
}
```

```css
.alert-\"hello-world\" {
    --color: #ff0000;
}
```

jannis-baum (approved):

PR #277, opened 2025-12-08
fix(#242): allow nested alerts

### LINKED ISSUE #242: Allow nested alerts
Followup #233 

I left nested alerts as a followup issue:

```
> [!NOTE]
> works
>> [!NOTE]
>> to be implemented
```

Image

This is something that Obsidian supports, but GitHub does not.

IMO, it makes no sense not to be supported.

tuurep:
Closes #242 

image

Fix was way simpler than I expected

A lot was referenced from md-it plugin https://github.com/ryanisaacg/markdown-it-callouts

Marked as Draft because I want to fix one thing:

In Obsidian, callouts can't be nested into (regular) blockquotes:

image

If we similarly disallow that, we won't have to bother with ambiguous cases such as this one:

```
>>> [!NOTE]
```

Currently, this renders as a single Alert:

image

In Obsidian, renders as a regular nested blockquote:

image

The Obsidian case is much clearer. I'd say let's make it so Blockquotes can't nest Alerts in them.

### DIFF
--- a/src/parser/alerts.ts
+++ b/src/parser/alerts.ts
@@ -115,29 +115,41 @@ const MarkdownItAlerts = (md: MarkdownIt) => {
         for (let i = 0; i < tokens.length; i++) {
             if (tokens[i].type !== 'blockquote_open') continue;
 
-            const open = tokens[i];
             const start = i;
+            const open = tokens[i];
+
             const end = findBlockquoteClose(tokens, start);
-
             if (end === null) continue;
-
             const close = tokens[end];
 
             // Get the first inline token, consists of:
             // 1. title line e.g. [!marker] Optional title
             // 2. the first paragraph in the alert body
-            let firstBlock;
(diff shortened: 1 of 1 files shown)


jannis-baum:
I finally got to take a proper look at this, sorry it took so long! Looks good, feel free to merge it! :)

> This is done! It would be possible to get all the `blockquote_open`s and `blockquote_close`s in a single pass instead of looping again for each `open`, which would be more performant but:
> 
> 1. Seemed like it would complicate the edge case checks
> 2. Nesting alerts is a rare case and the more you nest them, the worse of a document you're actually writing
> 
> If there's a simple way though, would be worth it.

I would argue that simplicity is more valuable than the performance here since

1. this is "just" a markdown-it plugin part of Vivify and it's already not super easy to understand so making it more complicated hurts maintainability
2. if you have so many alerts that you notice a performance issue with the current implementation then your document probably sucks haha

PR #281, opened 2026-01-08
Upgrade dependencies

### LINKED ISSUE #280: Upgrade dependencies

jannis-baum:
Close #280

Doesn't seem like anything broke. Turns out `yarn dev` was already broken before though ever since we introduced the intentionally circular symlinks😂 So I fixed that as well, first commit


tuurep (approved):
Yeah nice catch on the `yarn dev` breakage, confirmed it happens on main and fixed here

Yep doesn't seem like anything else broke on a quick try

PR #285, opened 2026-02-01
mdFilePattern instead or additionally to mdExtensions

### LINKED ISSUE #284: mdFilePattern instead or additionally to mdExtensions
What I would like to suggest is a new configuration option that will make the filetype detection of Vivify a little bit more configurable. I guess one could call it `mdFilePattern` such that it will enable the user to use a regular expression pattern in the determination what should be rendered as Markdown. Much like the filetype detection in Neovim with its `pattern`-argument of [`vim.filetype.add()`](https://neovim.io/doc/user/lua.html#vim.filetype.add()) just that it is fixed to the filetype markdown

Where could this be useful?
A limitation arises for me when I want to use Vivify together with [`calcurse`](https://github.com/lfos/calcurse) which is a handy text-based calendar. You can have notes for every item you can put into `calcurse`. I edit those notes with `nvim` and would like to use `vivify.vim` to get my note rendered in Markdown on side :) BUT the notes, for the time of the edit, exist temporarily as `/tmp/calcurse-note.Vf21ma` (the file suffix are 6 alphanumeric characters). With `mdFilePattern` I could solve the filedetection just as I did in `nvim` by providing the a pattern for the whole path of the file.

jannis-baum:
Close #284

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -58,8 +58,8 @@ keys:
   skipped. To test your expressions you can e.g. run the following in your
   browser console.
 
-  ```json
-  new RegExp("^my_[\\w]+$").test("my_path")
+  ```js
+  new RegExp("^.*/calcurse-note\\.[a-zA-Z0-9]{6}$").test("/tmp/calcurse-note.Vf21ma")
   ```
 
   By default there are no patterns set.
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -49,10 +49,20 @@ keys:
   code highlighting if available. The default Markdown extensions are
   `['markdown', 'md', 'mdown', 'mdwn', 'mkd', 'mkdn']`
 - **`"mdFilePatterns"`**\
-  An array of regular expression strings. Paths that match any of these regular
+  An array of [JavaScript regular
+  expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions)
+  *strings* that are passed into the
+  [`RegExp`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp)
+  constructor to build the expression. Paths that match any of these regular
   expressions will also be rendered as Markdown. Invalid expressions are
-  skipped. Note that expressions have to be escaped correctly for the JSON
-  format. By default there are no patterns set.
+  skipped. To test your expressions you can e.g. run the following in your
+  browser console.
+
+  ```json
+  new RegExp("^my_[\\w]+$").test("my_path")
+  ```
+
+  By default there are no patterns set.
 - **`"preferHomeTilde"`**\
   Prefer using `~` as a placeholder for your home directory in URLs as well as
   the `components` for `"pageTitle"` (default is `true`)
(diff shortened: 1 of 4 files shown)


docs/customization.md
```diff
@@ -48,6 +48,11 @@ keys:
   files (except for Jupyter Notebooks) will be displayed as monospaced text with
   code highlighting if available. The default Markdown extensions are
   `['markdown', 'md', 'mdown', 'mdwn', 'mkd', 'mkdn']`
+- **`"mdFilePatterns"`**\
+  An array of regular expression strings. Paths that match any of these regular
+  expressions will also be rendered as Markdown. Invalid expressions are
+  skipped. Note that expressions have to be escaped correctly for the JSON
+  format. By default there are no patterns set.
```
tuurep: Considerations from https://github.com/jannis-baum/Vivify/issues/284#issuecomment-3830965234

- info on regex flavor
    - (question: does any other option allow regex?)

As for escaping JSON, is it basically that you have to escape `"` and `\` and that's about it?

tuurep: I guess this is a good reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions#writing_a_regular_expression_pattern

And JSON makes it so we have to do double backslashes. Is this right?

tuurep: I was thinking this could use a practical example, what about adding the

```
"mdFilePatterns": ["^.*/calcurse-note\\.[a-zA-Z0-9]{6}$"]
```

example (https://github.com/jannis-baum/Vivify/issues/284#issuecomment-3830707384) to the docs? It's a really good one.

jannis-baum: Yeah I think that's it. The thing is that you'd usually write the pattern literal as `/pattern/` in JavaScript (i.e. without quotes like string literals) and that makes the escaping a lot easier

jannis-baum: I think we replied at the same time haha, I missed your idea of adding the calcurse thing. We could of course do that as well, do you think that'd be better?

tuurep: Hmm. Yeah, I think the calcurse example is very fitting, as it even showcases a double backslash.

the console example was useful too, I actually gave that a try. That also requires a double backslash but could be only coincidental :thinking:

docs/customization.md
```diff
@@ -49,10 +49,20 @@ keys:
...
+  *strings* that are passed into the
+  [`RegExp`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp)
+  constructor to build the expression. Paths that match any of these regular
   expressions will also be rendered as Markdown. Invalid expressions are
-  skipped. Note that expressions have to be escaped correctly for the JSON
-  format. By default there are no patterns set.
+  skipped. To test your expressions you can e.g. run the following in your
+  browser console.
+
+  ```json
```
tuurep:
````diff
-```json
+```js
````

jannis-baum: got spot haha

docs/customization.md
jannis-baum: Alrighty, what do you think about it now?

jannis-baum: > That also requires a double backslash but could be only coincidental 🤔

The string in the JS console thing is the same as it would need to be in the JSON

tuurep (approved):
Yeah I think it's sufficient 👍 Thanks

PR #289, opened 2026-02-03
Windows support

pidgeon777:


Tweekism:
Hi @pidgeon777 

Thanks for all your work on this one, sorry it has taken so long to get back to you on this. I had to setup my Windows environment to even test it, then i finished work and had to do it all again at home. 

So far I got vivify-server.exe to build but then crashes when loading a file or doing a directory listing.

This is using your latest commit of 9074ec9

I'm about to head to dinner and will look some more afterwards, what exact version of node are you using?

image

jannis-baum:
Hello @pidgeon777 ! A big thank you also from my side! As I've said in the issue I don't have any windows setup at all so I'll rely on @Tweekism and maybe other people who have commented on the issue to test this.

Looking ahead a bit, would you be interested in taking over long-term support/maintenance of Vivify Windows? If so, do you have the messaging app Signal? We have a Vivify contributor group there and would then like to invite you to join so we have an easier means of communication that also doesn't ping people on GitHub.

pidgeon777:
Hello @Tweekism, I'm still improving the branch.

```
node --version
v25.0.0
```

@jannis-baum I would love it, the problem is time. I have one job + running a startup, and time is very limited. Still, I'm working on an updated branch. If there is any issue, I will look at it when possible.

Tweekism:
> @jannis-baum I would love it, the problem is time. I have one job + running a startup, and time is very limited. Still, I'm working on an updated branch. If there is any issue, I will look at it when possible.

Yeah we are all a bit like that at the moment.

So far this is what I get on my end:-
- The node SEA binary seems to compile ok, no errors.
- Running vivify-server.exe directly runs as expected and will terminate automatically after a few seconds (this is expected behavior)
- While vivify server is running, browsing to <http://localhost:31622/viewer/~> will give me a directory listing and I can navigate the file system and I can view .md files rendered as HTML. However I don't get any css stylesheeting (see my first screenshot for example)
- Running `viv.ps1 <somepath>` throws an error (see below)
- Running `viv.cmd <somepath>` seems to work ok
- vivify-server.exe won't stay running once a browser is connected to it (@jannis-baum might be better to explain the expected behavior here  😅)




## Error from PowerShell 7.5.4
```
Unhandled exception. System.Management.Automation.PSInvalidOperationException: There is no Runspace available to run scripts in this thread. You can provide one in the DefaultRunspace property of the System.Management.Automation.Runspaces.Runspace type. The script block you attempted to invoke was:
        if ($Even.ll
        }

   at System.Management.Automation.ScriptBlock.GetContextFromTLS()
   at System.Management.Automation.ScriptBlock.InvokeAsDelegateHelper(Object dollarUnder, Object dollarThis, Object[] args)
   at lambda_method72(Closure, Object, DataReceivedEventArgs)
   at System.Diagnostics.AsyncStreamReader.FlushMessageQueue(Boolean rethrowInNewThread)
--- End of stack trace from previous location ---
   at System.Threading.QueueUserWorkItemCallback.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()

[process exited with code 3762504530 (0xe0434352)]
You can now close this terminal with Ctrl+D, or press Enter to restart.
``` 

## Error from Windows PowerShell 5.1.26100.7462
```
PS C:\Users\jason> viv .

[process exited with code 2 (0x00000002)]
You can now close this terminal with Ctrl+D, or press Enter to restart.
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

Install the latest PowerShell for new features and improvements! https://aka.ms/PSWindows

```

jannis-baum:
> However I don't get any css stylesheeting (see my first screenshot for example)

This may be related to how static files are served. The way this works is

- the `static/` directory in the repo's root is zipped in the build process and included in the binary (zipped in [`Makefile`](https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/Makefile#L27-L30), included as asset in [`sea-config.json`](https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/sea-config.json#L4-L6))
- we then access this zip file at runtime via the Node SEA library and serve the file's contents in the `/static` route (see [`src/routes/static.ts`](https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/src/routes/static.ts))

If I had to guess where this breaks on Windows I'd say it's [here](https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/src/routes/static.ts#L28) because the paths might not be correct (e.g. classical `/` vs. `\`), maybe swapping the string interpolation out for a proper `path` could fix it

> vivify-server.exe won't stay running once a browser is connected to it (@jannis-baum might be better to explain the expected behavior here 😅)

**TLDR;** If the static routing (see above) doesn't work then no websocket connections are established meaning the server doesn't know about its clients meaning it times itself out and terminates. What follows is the long version including how the timeout and the communication between server and client work, this might be relevant once the static routing is fixed.

The server will terminate on its own after the timeout when no clients are connected to it. Clients connect to the server via websockets. The timeout kill happens via [these functions](https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/src/app.ts#L30-L39) that are called in [`src/sockets.ts`](https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/src/sockets.ts) when the last socket disconnects or the first socket connects.

To be sure that the websocket clients are active and working there is a ping/pong that happens every second ([ping](https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/src/sockets.ts#L87-L96), [pong](https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/src/sockets.ts#L41-L48)); if the client fails to pong then we consider it disconnected and when there are no more clients (also in the initial state before the first client has connected) the timeout starts.

The initial establishing of the connection works as follows and includes communication of which path the client is registered for which is needed for the API, e.g. for the Vim plugin:

1. The path is hardcoded into the HTML served by the viewer: https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/src/routes/viewer.ts#L29
2. The client opens a websocket on page load: https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/static/client.mjs#L28
3. The client sends the path to the server via the websocket when it opens: https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/static/client.mjs#L88-L90
4. The server receives this message and registers the client for that path: https://github.com/jannis-baum/Vivify/blob/0fee6620b201c9707e2aa73b4709428f2eba3e3f/src/sockets.ts#L56-L78

pidgeon777:
I made many improvements and modifications since my last commit. I have to find some time and check everything before a further push.

PR #293, opened 2026-04-01
Add a way to specify which browser vivify should open.

### LINKED ISSUE #291: Add a way to specify which browser vivify should open.
Usecase: I don't want to use vivify on my main browser with all the customization. I want to use it more on a lightweight browser or something similar.

I don't know typescript so i can't code it and don't how feasable it is. Maybe with some flags when we launch viv?

Nonetheless, thank you jannis for your project. I found it two days ago and it is exactly what I need. I sent a PR to nixpkgs to add vivify.vim to the repo. It should be done in some hours.

jannis-baum:
Close #291

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -115,8 +115,8 @@ keys:
   [here](https://www.npmjs.com/package/open#app). Examples:
 
   ```json
-  "browserOptions": { name: "firefox" }
-  "browserOptions": { name: "chromium", arguments: ["--incognito"] }
+  "browserOptions": { "name": "firefox" }
+  "browserOptions": { "name": "chromium", "arguments": ["--incognito"] }
   ```
 
 ### Reloading config
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -110,6 +110,14 @@ keys:
   ```
 
   For alerts customization, see [full explanation with examples](alerts.md).
+- **`"browserOptions"`**\
+  Options to specify which browser to open, including arguments, as specified
+  [here](https://www.npmjs.com/package/open#app). Examples:
+
+  ```json
+  "browserOptions": { name: "firefox" }
+  "browserOptions": { name: "chromium", arguments: ["--incognito"] }
+  ```
 
 ### Reloading config
 
--- a/loader.mjs
+++ b/loader.mjs
@@ -1,0 +1,4 @@
+import { register } from 'node:module';
+import { pathToFileURL } from 'node:url';
+
+register('ts-node/esm', pathToFileURL('./'));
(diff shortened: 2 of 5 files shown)


docs/customization.md
```diff
@@ -110,6 +110,14 @@ keys:
...
 
   For alerts customization, see [full explanation with examples](alerts.md).
+- **`"browserOptions"`**\
+  Options to specify which browser to open, including arguments, as specified
+  [here](https://www.npmjs.com/package/open#app). Examples:
+
+  ```json
+  "browserOptions": { name: "firefox" }
+  "browserOptions": { name: "chromium", arguments: ["--incognito"] }
+  ```
```
tuurep, line 120: Looking here a little bit, haven't found out about the failing tests in `yarn test`, but I could successfully change the browser just as long as "name" and "arguments" are quoted (cursed JSON again)


````suggestion
  ```json
  "browserOptions": { "name": "firefox" }
  "browserOptions": { "name": "chromium", "arguments": ["--incognito"] }
  ```
````

I tested it with `vivaldi-stable` btw, with the `--incognito` flag also, which worked.

**Comment by tuurep** [comment]
`yarn test` fails on `main` too (not related to this PR)

**Comment by tuurep** [comment]
A lead:

`yarn test` is this:

```
node --loader ts-node/esm --test tests/unit/cli.ts tests/unit/alerts.ts
```

It fails with something like:

```
node:internal/modules/run_main:107
    triggerUncaughtException(
    ^
[Object: null prototype] {
  Symbol(nodejs.util.inspect.custom): [Function: [nodejs.util.inspect.custom]]
}
```

But if you run the two test files individually without the `--test` flag, they work, i.e.:

- `node --loader ts-node/esm tests/unit/alerts.ts`
- `node --loader ts-node/esm tests/unit/cli.ts`

What would that tell us?

tuurep (changes requested):
Anyways, looks good apart from the faulty JSON in docs!

docs/customization.md
jannis-baum: every time!! haha

jannis-baum:
@tuurep can you reproduce `yarn test` failing locally? For me it runs without errors both on this branch as well as on main, outputs the following: [test-output.txt](https://github.com/user-attachments/files/26667152/output.txt)

jannis-baum:
Seems like the issue was caused by a node update and I still had the old version. Should be good now, let's merge #296 first and then this one afterwards (I already rebased on top of that one to check that everything works here as well)

tuurep (approved):
Ok, it works for me

PR #296, opened 2026-04-13
Fix TypeScript loader

### LINKED ISSUE #295: Fix TypeScript loader
`node --loader` was deprecated in 25 which broke the tests & dev setup

jannis-baum:
Close #295

`node --loader` was deprecated in Node 25 which broke the tests & dev setup. i tried to look for some convention to follow here but couldn't find anything, it's all a mess. the way it is now is whatever copilot came up with🙃

### DIFF
--- a/loader.mjs
+++ b/loader.mjs
@@ -1,0 +1,4 @@
+import { register } from 'node:module';
+import { pathToFileURL } from 'node:url';
+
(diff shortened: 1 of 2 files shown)


tuurep (approved):
Sounds concerning.. Confirmed the tests are fixed with this, and I don't have better suggestions

PR #301, opened 2026-08-03
feat: add Obsidian-style embeds and wiki-link display text

### LINKED ISSUE #300: Support for Obsidian-style embeds
Hi,

I would like to request support for Obsidian-style embeds in the Markdown viewer.

For images, this would mean rendering syntax such as:

```markdown
![[image.jpg]]
```

Ideally, the viewer would also support image sizing, for example:

```markdown
![[image.jpg|300]]
```

It would also be great if PDF embeds could be supported in a similar way, as they are in Obsidian:

```markdown
![[file.pdf]]
```

This would make the viewer much more compatible with Markdown files created or maintained in Obsidian, especially for users who already have notes containing many embedded images or PDFs.

Thanks for considering this feature.

binEpilo:
Close #300

This PR adds support for:

1. ![[file]] embeds for images, PDFs, video, audio, and Markdown notes
2. Optional embed sizing such as ![[image.png|300x200]]
3. Custom wiki-link text such as [[note.md|My Note]]

This was vibecoded and mainly added because I wanted the feature to be available for everyone to use.

Tests and documentation are included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ issue](https://github.com/jannis-baum/vivify/issues/new/choose) or
 - links to other files: [relative links like in
   GitHub](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#relative-links)
   as well as absolute file links
+- [Obsidian-style embeds](docs/embeds.md) (`![[file]]`) and wiki-links
+  (`[[file]]`)
 - [add styles, classes, ids or other attributes directly from
   Markdown](https://github.com/arve0/markdown-it-attrs?tab=readme-ov-file#examples)
 - table of contents with `[[toc]]`
--- a/docs/embeds.md
+++ b/docs/embeds.md
@@ -1,0 +1,86 @@
+# Embeds and wiki-links
+
+Vivify supports [Obsidian-style](https://help.obsidian.md/) wiki-links and
+embeds in addition to standard Markdown links.
+
+## Wiki-links
+
+With the syntax `[[page]]` you link to another file. If the link target has no
+file extension, a `.md` extension is assumed, matching Obsidian's behaviour:
+
+```md
+See [[other-note]], or [[other-note.md]] if you want an explicit extension.
+```
+
+To use custom link text, append a pipe and the desired text:
+
+```md
+[[other-note.md|My Other Note]]
+[[other-note|My Other Note]]
+```
+
+The left side is used for the link `href` (with `.md` appended when no
+extension is present), and the right side is shown as the link's text.
+
+## Embeds
+
+Prefixing a wiki-link with `!` turns it into an inline **embed** instead of a
+link:
+
+```md
+![[image.jpg]]
+```
+
+The path is resolved relative to the document you are viewing, just like regular
+relative links and images. Which HTML element is generated depends on the file
+type, detected from the file extension:
+
+- **Images** (`jpg`, `jpeg`, `png`, `gif`, `webp`, `svg`, `bmp`, `ico`, `avif`)
+  are embedded with an `<img>` element, just like a standard
+  `![](image.jpg)` reference.
+- **PDFs** (`pdf`) are embedded in an `<iframe>` rendered by the browser's
+  built-in PDF viewer.
+- **Videos** (`mp4`, `webm`, `ogv`, `mkv`, `mov`) are embedded with a
+  `<video controls>` player.
+- **Audio** (`mp3`, `wav`, `ogg`, `m4a`, `aac`, `flac`) is embedded with an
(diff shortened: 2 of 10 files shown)


jannis-baum (changes requested):
Hello, thanks for opening the PR! I finally had some time to look at it now.

Have you thoroughly tested all the features you are proposing? My first step with feature PRs like these is to look at the `markdown-additional.md` test file which you added examples to that do not exist in this repo, i.e. it is not testable. Once I created some of the files you are referencing there I found a few issues:

- the sizing only works for images, if you e.g. add it to an embedded Markdown viewer or embedded video and it won't resize because your CSS overwrites it
- embedding a PDF doesn't work and Vivify instead makes the browser download it; fixing this might be a bigger and separate topic so unless there is a very simple and clean solution it may be better to explicitly not support it yet
- there may be more, please test everything yourself

Overall the idea of the rendering test files is that they work as they are without needing to add other files. We explicitly do not want to add big files like images or videos to this repo to keep it small, those files will have to go into my [assets repo](https://github.com/jannis-baum/assets), you can open a PR there as well to add videos and images you reference from here. A PDF can be small enough to be feasible to add directly to this repo, I would propose e.g. exporting a small, simple Vivify-rendered document (see the [guide](https://github.com/jannis-baum/Vivify/blob/main/docs/pdfs.md)) to PDF and adding it so you can reference it from your test file.

Once you have made the rendering test document showcase all your features without requiring files that not everyone will have, please thoroughly test it and make sure everything you implemented works to the best of your knowledge, and that you have understood the code and would have implemented it like this yourself as well.

A common problem with vibecoding is that models love to implement things they should just take an existing library function for. This creates a messy codebase, unnecessary maintenance overhead, and bugs. Following are some notes from Claude on this in your PR.

---

### 1. Reuse: `escAttr` reimplements `md.utils.escapeHtml` — and does it incompletely

`embeds.ts` hand-rolls:

```ts
function escAttr(value: string): string {
    return value.replace(/["&]/g, (char) => (char === '"' ? '&quot;' : '&amp;'));
}
```

markdown-it already exposes `md.utils.escapeHtml` (escapes `& < > "`), which is the canonical helper — markdown-it's own docs use it. The from-scratch version omits `<` and `>`. With `html: true` this isn't a new XSS surface (an author can already write raw HTML — see `mermaid.ts`, which doesn't escape at all), so it's not a security regression, but it's exactly the "don't reimplement a library function" case you called out. Pass `md` into `renderEmbedHtml` and use `md.utils.escapeHtml`.

### 2. Minimality: several bits of dead/redundant code

- `VIDEO_MIME` / `AUDIO_MIME` maps: every extension in `VIDEO_EXTS`/`AUDIO_EXTS` has a corresponding MIME entry, so the `const typeAttr = type ? … : ''` fallback branch is unreachable. And for a single `<source>`, the browser sniffs the type from the extension anyway — Obsidian doesn't emit these. ~20 lines that could go, or the maps kept but the dead ternary dropped.
- `escAttr` is applied to `width`/`height`, which are guaranteed `\d+` by `parseSize`'s regex — escaping digits is a no-op.
- `parseSize` returns `height: match[2] ?? undefined` — `match[2]` is already `undefined` when the group doesn't match, so `?? undefined` is redundant.

### 3. Duplication with `wiki-links.ts` (convention/DRY)

`embeds.ts` re-implements the `[[`/`]]` scanning loop, the `|` pipe-split, and the extension detection that already live in `wiki-links.ts`. They've diverged too: `wiki-links.ts` uses `pbasename(x).indexOf('.') > -1` to detect an extension, while `embeds.ts` uses `pextname(x)`. Not wrong today, but two copies of the same intent drifting apart. A small shared helper (bracket-scan + target/pipe parse) would cut both files down and keep them in sync.

Relatedly, a minor convention point (not a blocker): the sibling `wiki-links.ts` builds output via markdown-it's token API (`state.push('link_open')` + `attrSet`), which auto-escapes; `embeds.ts` switches to raw `html_inline` strings. Raw-HTML **is** precedented here (`mermaid.ts`, `front-matter.ts`, `dot.ts`), so it's acceptable — but at least the image case could use an `image` token with `attrSet` to match the neighbour and get escaping for free.

jannis-baum:
Hello @binEpilo ! Just bumping this to ask if you're still going to work on it.

FYI, we added a [policy for using generative AI for contributions](https://github.com/jannis-baum/Vivify/blob/main/docs/ai-policy.md) now, might be helpful to read alongside our normal [contribution guideline](https://github.com/jannis-baum/Vivify/blob/main/docs/CONTRIBUTING.md) in this context :)

binEpilo:
now that obsidian allows editing files outside of one's vault, i honestly dont really know if obsidian support is even needed in Vivify at this point. therefore i dont think its worth the time it would take to work on it.

